PR #31, opened 2020-06-16
Unit-tests for leds.c module

TheSomeMan:


ojousima:
Could the  tests/leds/freertos/src/FreeRTOS_Posix be a submodule and can it be placed somewhere higher in folder hierarchy? It wouldn't make sense to keep copies of the FreeRTOS_Posix under each tested components, and having it as a submodule would keep the origin and license of code clearer. 

How do I run the unit tests and produce a report of results and coverage?

TheSomeMan:
> Thank you for the contribution, please confirm BSD-3-Clause licensing is ok for you.
yes, BSD-3-Clause is ok for me

PR #32, opened 2020-06-17
Refactor uart

ojousima:


main/ethernet.c
```diff
@@ -222,6 +222,13 @@ void ethernet_init()
     esp_eth_phy_t * phy = esp_eth_phy_new_lan8720 (&phy_config);
     esp_eth_config_t config = ETH_DEFAULT_CONFIG (mac, phy);
     esp_eth_handle_t eth_handle = NULL;
-    ESP_ERROR_CHECK (esp_eth_driver_install (&config, &eth_handle));
-    ESP_ERROR_CHECK (esp_eth_start (eth_handle));
+    esp_err_t err_code = esp_eth_driver_install (&config, &eth_handle);
+    if(ESP_OK == err_code)
+    {
+      esp_eth_start (eth_handle);
```
TheSomeMan, line 228: The returned error code is ignored. Maybe it's worth writing a message to the log?

main/uart.c
```diff
@@ -15,321 +15,451 @@
...
+int uart_send_data (const char * logName, const char * data)
+{
+    char buf[100] = { 0 };
+    const int len = strlen (data);
 
-	const int txBytes = uart_write_bytes(UART_NUM_1, data, len);
-	ESP_LOGI(logName, "Wrote to uart %d bytes, 0x%s", txBytes, buf);
+    for (int i = 0; i < len; i++)
+    {
+        sprintf (buf + (2 * i), "%02x", data[i]);
```
TheSomeMan, line 97: sprintf is very unsafe here, it can cause stack damaging if the length of data will be bigger than 49 bytes.
Please consider using snprintf or ESP_LOG_BUFFER_HEXDUMP

main/uart.c
```diff
@@ -15,321 +15,451 @@
...
+    }
 
-	return txBytes;
+    const int txBytes = uart_write_bytes (UART_NUM_1, data, len);
+    ESP_LOGI (logName, "Wrote to uart %d bytes, 0x%s", txBytes, buf);
```
TheSomeMan, line 101:
```suggestion
    ESP_LOGI (logName, "Wrote to uart %d bytes:", txBytes);
    ESP_LOG_BUFFER_HEXDUMP(logName, buf, len, ESP_LOG_INFO);
```

main/uart.c
```diff
@@ -15,321 +15,451 @@
...
+    {
+#if 0
+            char data[10] = { 0 };
+
+        case SET_FILTER:
+        {
+            uint16_t company_id = * (uint16_t *) arg;
+            data[1] = NRF_CMD1_LEN;
+            data[2] = NRF_CMD1;
+            memcpy (&data[3], &company_id, 2);
```
TheSomeMan, line 118: this assignment is not portable, it will work differently on big-/little- endian architectures

main/uart.c
ojousima: It would be better to pass the length of data as an argument, just in case that NULL byte has ever to be sent

main/ethernet.c
ojousima: The following lines are 
```
    }
    else
    {
         ESP_LOGE (TAG, "Ethernet fail");
    }
```
Ideally this error would be stored somehow and printed out to factory testing system

PR #51, opened 2020-06-25
Fix GPIO num to match A1 hardware

ojousima:


ojousima:
@laurijamsa binaries can be found at https://jenkins.ruuvi.com/job/ruuvi_gateway_esp-PR/51/ . 
Desired behavior as I understand it is at https://docs.ruuvi.com/gw-esp32-firmware/gw-esp32-button .

PR #52, opened 2020-06-25
Fix #9: WiFi connection page is misleading

### LINKED ISSUE #9: WiFi connection page is misleading
After connection to internet is established, the hotspot is dropped and set doesn't need to reconnect to anything.
image

TheSomeMan:


ojousima:
The hotspot is supposed to turn off after configuration is complete, ideally once the acknowledgement of wifi connection established has been received by browser. If I'm reading this PR correctly, the hotspot remains active?

TheSomeMan:
> The hotspot is supposed to turn off after configuration is complete, ideally once the acknowledgement of wifi connection established has been received by browser. If I'm reading this PR correctly, the hotspot remains active?

The hotspot is not turned off at all (before this patch and after it).
The problem that this patch solves is the inability to see the completion status in browser.
I commented two lines where two freertos threads are removed (dns sever and http server). Removing threads is very dangerous operation because all allocated resources are still busy. Sockets remain open, some mutexes may be stayed in locked state if the thread was removed at some specific moment.
And I doubt that turning off the http server is what you want. I guess that hotspot should be disabled, but http should continue to work and it should be accessible via local network, right?

ojousima:
> I commented two lines where two freertos threads are removed (dns sever and http server). Removing threads is very dangerous operation because all allocated resources are still busy. Sockets remain open, some mutexes may be stayed in locked state if the thread was removed at some specific moment.

Ok, please remove the lines entirely rather than just commenting them out. 

> And I doubt that turning off the http server is what you want. I guess that hotspot should be disabled, but http should continue to work and it should be accessible via local network, right?

The original idea is to disable the configuration interface until someone comes to press the reset button, but leaving the gateway configurable might be better behavior as long as the configuration interface is password-protected and password isn't same default for every device.

Issue #54, opened 2020-06-27
UI for the gateway configuration


TheSomeMan:
[Example of wizard](http://webmanager.fi/website/ruuvi-gateway/?access_token=f0fb73b9bf27a350b3308e8fee1871cd)
[https://github.com/ruuvi/ruuvi.gwui.html](https://github.com/ruuvi/ruuvi.gwui.html)

PR #55, opened 2020-06-29
#54: Update submodule esp32-wifi-manager (new UI for the gateway configuration)

TheSomeMan:


ojousima:
Reported by Henri: After connection, the gateway doesn't send data out.

PR #56, opened 2020-07-04
#54: Update submodule esp32-wifi-manager (add support for WiFi disconnection and reconnection)

TheSomeMan:


ojousima:
Is this stale after #57? If I'm correct the updates were included and this pull request now points to older version of wifi-manager?

PR #98, opened 2020-08-06
Fix #63: Continuous Integration: Add code coverage analysis

### LINKED ISSUE #63: Continuous Integration: Add code coverage analysis
Report code coverage by unit tests and reject pull requests where less than 80 % of new code is unit tested. Can use any tool. 
Code which is tightly coupled to hardware or ESP32 stack can be ignored in unit tests.

TheSomeMan:


ojousima:
Jenkins could also be configured to run clang-format and fail build if there are changes, goal is to enforce the .clang-format style.

TheSomeMan:
> Jenkins could also be configured to run clang-format and fail build if there are changes, goal is to enforce the .clang-format style.

I suggest doing this is in separate pull-request, I've created #99 for this

ojousima:
Thanks

Issue #78, opened 2020-07-17
Feature: Status page to monitor gateway health


ojousima:
Let (authenticated?) users check basic statistics such as uptime, number of tags seen, bytes sent/received over WAN.

ojousima:
It would also be possible to have a human-readable status page and a machine-readable status page, for example in Prometheus format: https://prometheus.io/docs/instrumenting/exposition_formats/

TheSomeMan:
Closed in #100, #103

PR #100, opened 2020-08-10
Add initial version of /metrics http endpoint (#78)

Scrin:
I guess this is more of a proposal of a way to implement metrics for health monitoring on a status page (to be implemented) as well as using standard monitoring tools. I just happened to implement this method as a little practice while exploring the project setup and structure.

This implementation produces a plaintext metric result in [Prometheus exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/) for direct use with compatible monitoring tools (ie. Prometheus or Zabbix) or a status page served by the GW itself (to be implemented, the status page would call the metrics endpoint periodically and update itself accordingly). The http endpoint itself is implemented in: https://github.com/ruuvi/esp32-wifi-manager/pull/27

Currently the exposed metrics are just the number of received advertisements, device uptime, and heap memory statistics for different [memory capabilities](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html#memory-capabilities), but it's easy to add more metrics to the list.

### DIFF
--- a/main/includes/ruuvidongle.h
+++ b/main/includes/ruuvidongle.h
@@ -54,6 +54,11 @@ struct adv_report_table
     adv_report_t table[MAX_ADVS_TABLE];
 };
 
+typedef struct gw_metrics
+{
+    uint64_t received_advertisements;
+} gw_metrics_t;
+
 #define RUUVI_DONGLE_CONFIG_HEADER          (0xAABBU)
 #define RUUVI_DONGLE_CONFIG_FMT_VERSION     (0x0004U)
 
@@ -115,10 +120,12 @@ typedef enum nrf_command_t
 extern struct dongle_config m_dongle_config;
 extern EventGroupHandle_t status_bits;
 extern mac_address_str_t gw_mac_sta;
+extern gw_metrics_t gw_metrics;
 
 void mac_address_bin_init(mac_address_bin_t* p_mac, const uint8_t mac[6]);
 mac_address_str_t mac_address_to_str(const mac_address_bin_t* p_mac);
 char * ruuvi_get_conf_json();
+char * ruuvi_get_metrics();
 void settings_get_from_flash (struct dongle_config * dongle_config);
 void settings_print (struct dongle_config * config);
 int settings_clear_in_flash (void);
--- a/main/metrics.c
+++ b/main/metrics.c
@@ -1,0 +1,94 @@
+#include "ruuvidongle.h"
+#include "esp_log.h"
+#include "esp_heap_caps.h"
+
+#define BUF_LEN 2048
+#define METRICS_PREFIX "ruuvigw_"
+
+static const char TAG[] = "metrics";
+
+gw_metrics_t gw_metrics = {};
+
+size_t get_total_free_bytes(const uint32_t caps)
+{
+    multi_heap_info_t x;
+    heap_caps_get_info(&x, caps);
+    return x.total_free_bytes;
+}
+
+// See:
+// https://prometheus.io/docs/instrumenting/writing_exporters/
+// https://prometheus.io/docs/instrumenting/exposition_formats/
+
+int gen_metrics(char* buf, size_t limit)
+{
+    return snprintf(buf, limit,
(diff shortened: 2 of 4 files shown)


ojousima:

ojousima:
Add to whitelist

main/includes/ruuvidongle.h
```diff
@@ -54,6 +54,11 @@ struct adv_report_table
     adv_report_t table[MAX_ADVS_TABLE];
 };
 
+typedef struct gw_metrics
+{
+    long long received_advertisements;
```
ojousima, line 59: uint64_t?

main/metrics.c
```diff
@@ -0,0 +1,94 @@
...
+#include "esp_heap_caps.h"
+
+#define BUF_LEN 2048
+#define METRICS_PREFIX "ruuvigw_"
+
+static const char TAG[] = "metrics";
+
+gw_metrics_t gw_metrics = {};
+
+size_t get_total_free_bytes(uint32_t caps)
```
ojousima, line 12: `size_t get_total_free_bytes(const uint32_t caps)`?

main/metrics.c
```diff
@@ -0,0 +1,94 @@
...
+    (size_t) heap_caps_get_largest_free_block(MALLOC_CAP_PID6),
+    (size_t) heap_caps_get_largest_free_block(MALLOC_CAP_PID7),
+    (size_t) heap_caps_get_largest_free_block(MALLOC_CAP_SPIRAM),
+    (size_t) heap_caps_get_largest_free_block(MALLOC_CAP_INTERNAL),
+    (size_t) heap_caps_get_largest_free_block(MALLOC_CAP_DEFAULT));
+}
+
+char* ruuvi_get_metrics()
+{
+    char *buf = malloc(BUF_LEN * sizeof(char));
```
ojousima, line 86: I think this is a bit dangerous as I can't see where `buf` is freed. Maybe allocate `buf` at callee and pass the pointer + size as parameter?

Scrin: The buf is freed in http_server.c (see the other PR) in the same way as ruuvi.json is freed there after copying it into the http response

main/metrics.c
```diff
@@ -0,0 +1,94 @@
...
+    heap_caps_get_largest_free_block(MALLOC_CAP_DEFAULT));
+}
+
+char* ruuvi_get_metrics()
+{
+    char *buf = malloc(BUF_LEN * sizeof(char));
+    int size = gen_metrics(buf, BUF_LEN);
+    if (size >= BUF_LEN) {
+        ESP_LOGW (TAG, "Initial buffer size of %d for metrics is insufficient, needed %d! Consider increasing BUF_LEN in metrics.c", BUF_LEN, size);
+        realloc(buf,(size + 1) * sizeof(char));
```
ojousima, line 90: I think realloc may move memory to a new location if the existing pointer cannot be expanded to fit new size and free the old pointer. New value of buffer should be stored and used.

main/metrics.c
ojousima: Ok, @TheSomeMan has final say here

ojousima:
Thank you for the contribution, we'll review it in details as soon as possible. 

I just noticed that repository doesn't have a license file online yet. Is BSD-3 license ok for you @Scrin?

main/metrics.c
Scrin: Good catch, I forgot realloc doesn't (necessarily) just modify the original one. I'll fix this when the decision is done where to allocate this. I think it's good to have a consistent way of doing things, but indeed it would be clearer to do both the allocation and freeing on the same side.

Scrin:
BSD-3 is ok for me

main/metrics.c
TheSomeMan: it's commonly used approach in this project, so I think it's ok for now, but in future we can refactor it.

main/metrics.c
TheSomeMan: there is a better way - calculate the required size for the buffer, then allocate it. I'll do a refactoring.

PR #109, opened 2020-08-24
#101: Update esp32-wifi-manager: fix security vulnerability - the use of insecure "sprintf" function

TheSomeMan:
json lib refactoring, unit-tests


ojousima (approved):
Thanks

PR #112, opened 2020-08-27
#101: Fix vulnerability (the use of this insecure "sprintf" function), update esp32-wifi-manager (fix security hotspots)

TheSomeMan:
Fix vulnerability: remove the use of this insecure "sprintf" function.
Update esp32-wifi-manager: fix security hotspots: refactoring of unsafe strcpy and strncpy usage.


ojousima (approved):
Thanks

PR #132, opened 2020-10-22
README.md: Fix SWD connection description

TheSomeMan:


ojousima (commented):
I'm having some issue on my setup with this wiring, I'll test it out.

PR #138, opened 2020-11-01
Fix #70: Refactoring: move the gateway dependent code from esp32-wifi-manager submodule to the main repo

### LINKED ISSUE #70: Refactoring: move the gateway dependent code from esp32-wifi-manager submodule to the main repo
http_server.c from esp32-wifi-manager submodule contains code which is dependent on ruuvidongle.h from this repo, which leads to cycle in dependencies.
Need to refactor both components and make esp32-wifi-manager completely independent.

TheSomeMan:


ojousima (approved):
Thanks

PR #139, opened 2020-11-03
#101 Refactoring, fix code smells and bugs found by sonarcloud

TheSomeMan:


main/adv_post.c
```diff
@@ -299,5 +295,8 @@ adv_post_init(void)
     adv_reports.num_of_advs = 0;
     api_callbacks_reg((void *)&adv_callback_func_tbl);
     const uint32_t stack_size = 1024U * 4U;
-    xTaskCreate(&adv_post_task, "adv_post_task", stack_size, NULL, 1, NULL);
+    if (!os_task_create(&adv_post_task, "adv_post_task", stack_size, NULL, 1, NULL))
+    {
+        LOG_ERR("Can't create thread");
```
ojousima, line 300: Some of the task creation failures are fatal, for example failing here means gateway is functional. What would be appropriate action to take? Reboot? Erase configured settings and reboot? Blink error through LEDs and let user figure it out?

ojousima:
Let's take a closer look at actions on task creation failure. On nRF52 firmware we have a flag for fatal error in error code and a error handler which will reboot the MCU in case of a fatal error.

Another approach would be to have a watchdog which is reboots the gateway e.g. if no sensor data is sent to backend and a response code 200 is received for an hour.

ojousima (approved):
Thanks, let's look at the task creation error handling separately

PR #157, opened 2020-12-20
Fix #150: The time synchronization thread does not handle WiFi/Ethernet disconnection 

### LINKED ISSUE #150: The time synchronization thread does not handle WiFi/Ethernet disconnection

TheSomeMan:


ojousima (approved):
Thanks

PR #159, opened 2020-12-23
Fix #158: Increase the number of NTP servers from 1 to 4

### LINKED ISSUE #158: Increase the number of NTP servers from 1 to 4

TheSomeMan:


ojousima (approved):
Thanks

PR #169, opened 2021-01-13
Fix #167: dns_server and http_server are not properly stopped

### LINKED ISSUE #167: dns_server and http_server are not properly stopped

TheSomeMan:


ojousima (approved):
Thanks

PR #174, opened 2021-01-24
Include num_of_advs in log message

DG12:
Was
ESP_LOGI(TAG, "Advertisements in table:);
is
ESP_LOGI(TAG, "Advertisements in table: %d", p_reports->num_of_advs); // added space


ojousima:
ok to test

Issue #25, opened 2020-02-24
Handle lot of tags more gracefully


ojousima:
e.g. Store data into hash map by tag MAC and send one datapoint per each tag rather than saturating the rambuffer and dropping new advertisements.

PR #176, opened 2021-02-07
#25: Handle lot of tags more gracefully

TheSomeMan:


main/adv_table.c
```diff
@@ -0,0 +1,174 @@
...
+
+ADV_TABLE_STATIC
+uint32_t
+adv_report_calc_hash(const mac_address_bin_t *const p_mac)
+{
+    uint32_t     hash_val           = 0;
+    const size_t mac_addr_half_size = sizeof(p_mac->mac) / sizeof(p_mac->mac[0]) / 2;
+    for (uint32_t i = 0; i < mac_addr_half_size; ++i)
+    {
+        hash_val |= ((uint32_t)p_mac->mac[i] ^ (uint32_t)p_mac->mac[i + mac_addr_half_size]) << (i * CHAR_BIT);
```
ojousima, line 66: Is there some source which explains this hash function?

TheSomeMan: It's just XOR-based hash calculation - one of the simplest and effective method for hashing.
For example, here is comparison of various hashing functions (XOR, checksum, CRC): https://www.cs.wustl.edu/~jain/papers/ftp/hash_iee.pdf
If this code 24-bit has value is calculated as XOR between two 24-bit parts of MAC address.
Since the higher 24-bits of bluetooth MAC address are fixed for a vendor, so in most practical cases (for devices of one vendor) all hashes  will have unique values.
This 24-bit hash value is divided to some prime number (101 in this case) to calculate an index in the hash table.

ojousima (approved):
Thanks

PR #178, opened 2021-02-15
Add hotfix for A2 GPIO

ojousima:
Just getting Jenkins to compile binaries to test the SWD update on A2 hardware


ojousima:
Works, but should be cleaned up or closed and implemented cleanly from the start.

Issue #137, opened 2020-10-31
Add an HTTP-endpoint to be able to poll the latest data from the gateway using HTTP-GET


TheSomeMan:
https://f.ruuvi.com/t/ruuvi-sensors-ruuvi-gateway-home-assistant-easy/
https://f.ruuvi.com/t/ruuvi-sensors-ruuvi-gateway-home-assistant-easy/4445/5

io53:
@TheSomeMan do you have any estimate when we could expect this feature to be implemented?

ojousima:
It's a question of priorities, I think this could be assigned to milestone 1.6

TheSomeMan:
@io53, I agree with @ojousima, it's a matter of priorities. 
Now I'm working on v1.3.1 and we have several versions in queue before v1.6: v1.3.2, v1.3.3, v1.4 and v1.5.
So, it's quite hard to predict now when it will be implemented.
But it's not a very difficult task, I think it can be implemented in 3-4 days (with unit-tests).
Maybe it's worth moving this task to v1.3.2 since there is related task #25 there, what do you think @ojousima?

io53:
As for the format of the response, will it be about the same as the HTTP post body it currently sends?
The feature does not have to be implemented in the gateway for development of the HA integration to start as long as we have an idea of how the api will work :).

TheSomeMan:
@io53 what do you think what format is more convenient?
we can send data in json format like in HTTP post body or in plain text like in "/metrics" endpoint

io53:
Json would be a lot easier to work with for sure.

TheSomeMan:
New HTTP-endpoint:
/history[&time=<interval_seconds>]

PR #192, opened 2021-03-13
#137: Add "/history" HTTP-endpoint to be able to poll the latest data from the gateway using HTTP-GET

TheSomeMan:


ojousima (approved):
Thanks

PR #199, opened 2021-03-18
Fix #177: Change hostname of the gateway to "RuuviGatewayXXXX" to make it easier to identify it on the network

### LINKED ISSUE #177: Change hostname of the gateway to "RuuviGatewayXXXX" to make it easier to identify it on the network
currently hostname is "espressif"

TheSomeMan:


ojousima (approved):
Thanks

PR #203, opened 2021-03-22
Close #202: Perform software reset for nRF52 during SWD initialization

### LINKED ISSUE #202: Perform software reset for nRF52 during SWD initialization

TheSomeMan:


ojousima (approved):
Thanks

PR #208, opened 2021-03-24
Fix #207: History API query string expects "&" instead of "?"

### LINKED ISSUE #207: History API query string expects "&" instead of "?"
The api only works when the query string separator is "&" eg. http://IP/history&time=30
Proper separator is "?" so this following is the correct way http://IP/history?time=30

TheSomeMan:


ojousima (approved):
Thanks

PR #221, opened 2021-04-04
Fix #10: No indication on connectivity failures

### LINKED ISSUE #10: No indication on connectivity failures
If the ESP crashes on configuration save, there's only a brief moment of LED being dark as an indication of failed configuration. 

User can proceed to configure the WiFi unaware that his configuration was never saved.

Same applies to mistyped URLs

TheSomeMan:


ojousima (approved):
Thanks

PR #223, opened 2021-04-05
Fix #216: Short press on the Configure button does not activate WiFi AP

### LINKED ISSUE #216: Short press on the Configure button does not activate WiFi AP

TheSomeMan:


ojousima (approved):
Thanks

PR #235, opened 2021-05-18
Close #234: Implement hardware reset for nRF52 on Gateway rev.A2 (requires removing R15 resistor from the board)

### LINKED ISSUE #234: Implement hardware reset for nRF52 on Gateway rev.A2 (requires removing R15 resistor from the board)

TheSomeMan:


ojousima (approved):
Thanks

Issue #60, opened 2020-07-08
ESP32 firmware update over the internet


ojousima:
The ESP32 should be able to download update image and run update over hotspot configuration. No automatic updates over the internet without user interaction.

ojousima:
The update page could be in a separate tab or something like that, and the update page could have human-readable status information as in #171. 

@sini-maaria do you have a chance to make a mock-up of page in near future?

PR #239, opened 2021-06-04
#60: ESP32 firmware update over the internet

TheSomeMan:


ojousima (approved):
`esp_ota_ops_patched.c`  has some frees / allocates I'm not comfortable with, but if it comes from Espressif I'm sure they have enough experience to be sure it doesn't leak memory. 

Ok to merge by me, thanks

PR #244, opened 2021-06-09
Fix #243: Project version does not contain "-dirty" suffix if git repo contains modified files

### LINKED ISSUE #243: Project version does not contain "-dirty" suffix if git repo contains modified files

TheSomeMan:


ojousima (approved):
Thanks

PR #261, opened 2021-07-05
Close #111: Sign sent BLE data sent with tag ID

### LINKED ISSUE #111: Sign sent BLE data sent with tag ID
Depends on #110, gateway data should be signed with device ID somehow. For example if data sent as JSON is 
```
{
  "data": {
    "coordinates":  "",
    "timestamp":  1595504641,
    "gwmac":  "240ac4e32eb4",
    "tags": {. . .
    }
  }
}
```
the signed data could be
```
{
  "data": {
    "coordinates":  "",
    "timestamp":  1595504641,
    "gwmac":  "240ac4e32eb4",
    "tags": {. . .
    }
  },
  "signature":$SIGNATURE
}
```
where value of signature is SHA256(data, device_id).

The goal is to prevent data spoofing.

TheSomeMan:


main/http.c
```diff
@@ -106,9 +114,15 @@ http_send(const char *const p_msg)
         return false;
     }
 
-    esp_http_client_set_post_field(http_handle, p_msg, strlen(p_msg));
+    esp_http_client_set_post_field(http_handle, p_msg, (int)strlen(p_msg));
     esp_http_client_set_header(http_handle, "Content-Type", "application/json");
 
+    const hmac_sha256_str_t hmac_sha256 = hmac_sha256_calc(p_msg);
+    if ('\0' != hmac_sha256.buf[0])
```
ojousima, line 121: Does something guarantee that first byte of sha256 buffer is never 0 on successful calculation? It's my understanding that all values 0 ... 255 are equally likely, which would cause this to fail 1/255 times

main/adv_post.c
```diff
@@ -289,6 +304,7 @@ adv_post_task(void)
 void
 adv_post_init(void)
 {
+    g_adv_post_nonce = esp_random();
```
ojousima, line 307: I think the nonce is supposed to be updated for each request and sent along in the header. This prevents replay attacks

ojousima (commented):
Let's check the comments before merging

main/adv_post.c
ojousima: Nonce gets incremented after post, ok

main/http.c
TheSomeMan: it's a string, so if it's not empty it means that it contains hex-encoded HMAC_SHA256.
In any case, I've refactored this part to avoid confusion.

ojousima (approved):
Thanks

PR #264, opened 2021-07-11
Fix #255: Guru Meditation Error after accepting a network connection in rare cases

### LINKED ISSUE #255: v1.4.0-4 Guru Meditation Error on HTTP request  when the network connection is lost before handling the request

TheSomeMan:


ojousima (approved):
Thanks

PR #266, opened 2021-07-11
Fix #265: Sometimes there is a long delay (about 30 seconds) between connecting to WiFi and opening the Gateway configuration page

### LINKED ISSUE #265: Sometimes there is a long delay (about 30 seconds) between connecting to WiFi and opening the Gateway configuration page

TheSomeMan:


ojousima (approved):
Thanks

PR #268, opened 2021-07-11
Fix #267: UI: Show a notification when WiFi connection to the Gateway is lost

### LINKED ISSUE #267: UI: Show a notification when WiFi connection to the Gateway is lost

TheSomeMan:


ojousima (approved):
Thanks

PR #269, opened 2021-07-11
Fix #262: sdkconfig: increase CPU freq from 160 to 240 MHz, increase FreeRTOS freq from 100 to 1000 Hz (to solve the problem with frequent WiFi disconnects during firmware upgrades)

### LINKED ISSUE #262: WiFi is often lost during firmware upgrades

TheSomeMan:


ojousima (approved):
Thanks

PR #272, opened 2021-07-14
Fix #271: When upgrading the gateway firmware, the previous firmware is flashed on the nRF52, and after the reboot the correct one is flashed

### LINKED ISSUE #271: When upgrading the gateway firmware, the previous firmware is flashed on the nRF52, and after the reboot the correct one is flashed

TheSomeMan:


ojousima (approved):
Thanks

PR #275, opened 2021-07-14
Fix #274: WiFi AP is not stopped after configuring the gateway to use Ethernet

### LINKED ISSUE #274: WiFi AP is not stopped after configuring the gateway to use Ethernet

TheSomeMan:


ojousima (approved):
Thanks

Issue #108, opened 2020-08-19
Cloud configuration


ojousima:
Configuration for gateway transmission interval, nRF52 settings etc over cloud backend.

ojousima:
One potential implementation: When gateway sends data via HTTP, the reply has code letting gateway to know there is new configuration and configuration data is in reply.

Muhwu:
The current implementation sends the desired response rate back in a HTTP header `x-ruuvi-gateway-rate` header which is equivalent to the `GATEWAY_SEND_RATE` variable in the lambda configuration.

This value can be translated to `Please send a request every X seconds`. This can be changed to be a part of the API response if it better suits the purpose, but given that returning the ratio in the header is a low cost operation both in terms of transfer and adding it _and_ because we can't rely on the recipient receiving it 100%, perhaps it'd be best if the gateway used whatever value was returned.

Later on, we can make this configurable by gateway if we so wish.

PR #276, opened 2021-07-14
#108: Update ruuvi.gwui.html submodule: http_server_auth.py: add support for x-ruuvi-gateway-rate header

TheSomeMan:


ojousima (approved):
Thanks

PR #277, opened 2021-07-18
Fix #262: UI: Perform requests for ap.json only when the network list page is displayed and do it in sync manner (in order to prevent WiFi connection lost)

### LINKED ISSUE #262: WiFi is often lost during firmware upgrades

TheSomeMan:


ojousima (approved):
Thanks

PR #279, opened 2021-07-19
Close #278: Starting HTTP-server and displaying the nRF52 firmware update progress during the update process which is performed just after the reboot

### LINKED ISSUE #278: Starting HTTP-server and displaying the nRF52 firmware update progress during the update process which is performed just after the reboot

TheSomeMan:


ojousima (approved):
Thanks

PR #284, opened 2021-07-25
Fix #283: UI: When accessing from LAN to a gateway configured with auth_type=deny there is no redirection to auth.html

### LINKED ISSUE #283: UI: When accessing from LAN to a gateway configured with auth_type=deny there is no redirection to auth.html

TheSomeMan:


ojousima (approved):
Thanks

Issue #290, opened 2021-08-05
Configuration UI redesign


laurijamsa:
Make the layout fully responsive so that it'll work well on any display size (mobile, tablet, desktop).

Translation: Several strings have changed and no Finnish translations available for all of them. In this case, you can just duplicate the English strings and after the html is otherwise ready, someone will translate if to Finnish at once and we'll also double check the English grammar once all is otherwise ready.

PDF: 
[qw-config-v1-final.pdf](https://github.com/ruuvi/ruuvi.gateway_esp.c/files/6937005/qw-config-v1-final.pdf)

![image](https://user-images.githubusercontent.com/1311912/128314117-a3f80013-8242-4ef6-b8d7-4e7749bf2deb.png)

Logo: 
[ruuvi-logo-2021.svg.zip](https://github.com/ruuvi/ruuvi.gateway_esp.c/files/6937016/ruuvi-logo-2021.svg.zip)

markoaamunkajo:
Issues #297 #298 #299 #301 opened for reference, might not be valid for new UI

laurijamsa:
This is completed. From now on, let's open UI related issues on the correct repository.

PR #302, opened 2021-08-22
#290: Configuration UI redesign

TheSomeMan:


ojousima (commented):
Let's go through the reset in Tuesday 24.8 meeting

ojousima (approved):
Thanks

PR #310, opened 2021-08-25
#294: LED_TASK: Change network_indication so that red led is off when connection is ok

TheSomeMan:


ojousima:
How does the LED behave if WiFi / Ethernet connection is ok, but data cannot be sent via MQTT / HTTP because of remote server down / authentication issue etc?

TheSomeMan:
> How does the LED behave if WiFi / Ethernet connection is ok, but data cannot be sent via MQTT / HTTP because of remote server down / authentication issue etc?

in http_send_advs:
```
    if (http_send(json_str.p_str))
    {
        leds_indication_on_network_ok();
    }
    else
    {
        leds_indication_network_no_connection();
    }
```

in mqtt_event_handler:
```
        case MQTT_EVENT_CONNECTED:
            LOG_INFO("MQTT_EVENT_CONNECTED");
            xEventGroupSetBits(status_bits, MQTT_CONNECTED_BIT);
            mqtt_publish_connect();
/* --> */ leds_indication_on_network_ok();
            break;

        case MQTT_EVENT_DISCONNECTED:
            LOG_INFO("MQTT_EVENT_DISCONNECTED");
            xEventGroupClearBits(status_bits, MQTT_CONNECTED_BIT);
/* --> */ leds_indication_network_no_connection();
            break;
```

ojousima (approved):
Thanks

PR #311, opened 2021-08-25
#290: Update ruuvi.gwui.html: small fixes on pages "welcome", "ethernet_connection" and "software_update"

TheSomeMan:
Fix conflicts between "hidden" and "block-centered" on "page-ethernet_connection" and "page-software_update"; replace the button text "Begin" with "LET'S GET STARTED!"


ojousima (approved):
Thanks

PR #333, opened 2021-08-29
#325: Fix: The firmware update check runs every 40 minutes even after a successful attempt (should be 12 hours)

TheSomeMan:


ojousima (approved):
Thanks

Issue #326, opened 2021-08-28
Set time sync mode to SNTP_SYNC_MODE_IMMED after reboot and switch to SNTP_SYNC_MODE_SMOOTH after first synchronization


TheSomeMan:

laurijamsa:
I can see this happening from the logs and so far I haven't seen bad responses from the server. But needs more testing.

PR #334, opened 2021-08-29
#326: Set time sync mode to SNTP_SYNC_MODE_IMMED after reboot and switch to SNTP_SYNC_MODE_SMOOTH after first synchronization

TheSomeMan:


ojousima (approved):
Thanks

Issue #327, opened 2021-08-28
Memory leak when checking for firmware updates if the latest version is already installed


TheSomeMan:

laurijamsa:
Log for reference:

```
E (50915376) emac_esp32: no mem for receive buffer
E (50916562) emac_esp32: no mem for receive buffer
E (50916568) emac_esp32: no mem for receive buffer
E (50916797) emac_esp32: no mem for receive buffer
E (50916797) emac_esp32: no mem for receive buffer
E (50916863) emac_esp32: no mem for receive buffer
E (50916863) emac_esp32: no mem for receive buffer
E (50916863) emac_esp32: no mem for receive buffer
E (50917163) emac_esp32: no mem for receive buffer
E (50917669) emac_esp32: no mem for receive buffer
E (50918720) emac_esp32: no mem for receive buffer
E (50919549) emac_esp32: no mem for receive buffer
E (50919743) emac_esp32: no mem for receive buffer
E (50919923) emac_esp32: no mem for receive buffer
E (50919923) emac_esp32: no mem for receive buffer
W (50920276) esp-tls: Failed to open new connection in specified timeout
E (50920276) TRANS_SSL: Failed to open a new connection
E (50920280) HTTP_CLIENT: Connection failed, sock < 0
E (50920284) http: [adv_post_task] ../main/http.c:157 {http_send}: HTTP POST to URL=https://network.ruuvi.com/record: request failed, err=28674 (ERRO)
I (50920298) LEDS: [adv_post_task] leds_indication_network_no_connection
I (50920305) LEDS: [adv_post_task] LED: Start blinking, interval: 200 ms, duty cycle: 50%
I (50922874) ruuvi_gateway: [main] free heap: 46432
I (50932874) ruuvi_gateway: [main] free heap: 46432
I (50942874) ruuvi_gateway: [main] free heap: 46432
I (50952874) ruuvi_gateway: [main] free heap: 46432
I (50962874) ruuvi_gateway: [main] free heap: 46432
I (50972874) ruuvi_gateway: [main] free heap: 46432
```

laurijamsa:
How could we test this without waiting x times 12h after a successful update to 1.8.1? I think this occurred also without first successfully auto updated to a latest release. If that's true, should we make a local build that has the "12h polling setting" something shorter but otherwise the same as the release? Or how?

TheSomeMan:
[...]

PR #335, opened 2021-08-29
#327: Fix: Memory leak when checking for firmware updates if the latest version is already installed

TheSomeMan:


ojousima (approved):
Thanks

Issue #328, opened 2021-08-28
v1.7.1: Guru Meditation Error when configuring the gateway using phone (during connecting to WiFi)


TheSomeMan:
![image](https://user-images.githubusercontent.com/4346875/131227907-3c747a6e-1671-4aa7-adec-37ee72e644f9.png)

```
W (111678) ADV_POST_TASK: [adv_post_task] Can't send, no network connection
I (111731) dns_server: [dns_server] Replying to DNS request for connectivitycheck.gstatic.com from 10.10.0.2
I (111734) dns_server: [dns_server] Replying to DNS request for www.google.com from 10.10.0.2
I (111753) dns_server: [dns_server] Replying to DNS request for www.google.com from 10.10.0.2
I (112879) ruuvi_gateway: [main] free heap: 127120
Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.
Core 0 register dump:
PC      : 0x400014fd  PS      : 0x00060430  A0      : 0x80184184  A1      : 0x3ffcdd60
A2      : 0x00000000  A3      : 0xfffffffc  A4      : 0x000000ff  A5      : 0x0000ff00
A6      : 0x00ff0000  A7      : 0xff000000  A8      : 0x00000000  A9      : 0x3ffcdd10
A10     : 0x00000000  A11     : 0x3ffcf152  A12     : 0x3ffcdfa4  A13     : 0x3ffae920
A14     : 0x00000000  A15     : 0x00000001  SAR     : 0x00000004  EXCCAUSE: 0x0000001c
EXCVADDR: 0x00000000  LBEG    : 0x400014fd  LEND    : 0x4000150d  LCOUNT  : 0xffffffff
ELF file SHA256: 596b9478ec6eb2a6
Backtrace: 0x400014fa:0x3ffcdd60 0x40184181:0x3ffcdd70 0x401857f9:0x3ffce080 0x400820b5:0x3ffce0b0 0x400820d9:0x3ffce0e0 0x400ef484:0x3ffce130 0x400efb04:0x3ffcf210 0x400fdbd3:0x3ffcf250 0x4008a7d1:0x3ffcf270
Rebooting...
ets Jul 29 2019 12:21:46
```

PR #336, opened 2021-08-29
#328: Fix: v1.7.1: Guru Meditation Error when configuring the gateway using phone (during connecting to WiFi)

TheSomeMan:


ojousima (approved):
Thanks

Issue #304, opened 2021-08-23
Connection to Gateway hotspot is lost while scanning Wi-Fi networks or trying to connect to one


laurijamsa:
![image](https://user-images.githubusercontent.com/1311912/130392983-b298a16b-7fa5-4ea7-a6e6-d794ccab2dcc.png)

hakunti:
![image](https://user-images.githubusercontent.com/25771106/130415815-9a2cf0aa-b630-4589-b78d-680d22c10fb4.png)

ojousima:
This was probably related to #303 and fixed in c18e1e1.  Closing for now

TheSomeMan:
@ojousima, @hakunti faced with another type of the problem which causes "Guru meditation error", now I'm trying to find the cause

hakunti:
With the latest built I was not able to have anymore same issues. There are small glitches with the next button but it's maybe loading the next page.

TheSomeMan:
yes, it was fixed in https://github.com/ruuvi/esp32-wifi-manager/issues/112

laurijamsa:
Still getting this on the build 345 when trying to connect to a Wi-Fi network.

![image](https://user-images.githubusercontent.com/1311912/130601480-5d7aa4c1-7f19-4423-b031-db3f05751d1b.png)

Log:

``` 

(22705) wifi_manager: [sys_evt] ../components/esp32-wifi-manager/src/wifi_manager.c:524 {wifi_manager_event_handler}: WE
D (22707) wifi_manager: [wifi_manager] ../components/esp32-wifi-manager/src/wifi_manager.c:626 {wifi_handle_ev_scan_done}E
I (22721) http_server: [http_server] ap.json: [{"ssid":"Secret Network 2G","chan":9,"rssi":-59,"auth":3},
{"ssid":"AndroidAP22BE","chan":6,"rssi":-90,"auth":3}
]

I (22735) http_server: [http_server] Json resp: code=200, content:
[{"ssid":"Secret Network 2G","chan":9,"rssi":-59,"auth":3},
{"ssid":"AndroidAP22BE","chan":6,"rssi":-90,"auth":3}
]

I (22838) http_server: [http_server] Request from 10.10.0.2 to 10.10.0.1: GET /status.json
I (22839) http_server: [http_server] GET /status.json
I (22842) http_server: [http_server] status.json: {}

I (22848) http_server: [http_server] Json resp: code=200, content:
{}

[...]

PR #343, opened 2021-08-31
#304: Implement Wi-Fi channel scanning one after another with a delay to give time to the main channel

TheSomeMan:


ojousima (approved):
Thanks

PR #345, opened 2021-09-01
#290: Update ruuvi.gwui.html - On macOS hotspot configuration window is scrolled automatically; Popup window "Connection to the Ruuvi Gateway is lost" looks not good

TheSomeMan:


ojousima (approved):
Thanks

PR #346, opened 2021-09-02
Close #315: UI: Move the authentication setting page to the basic flow

### LINKED ISSUE #315: Allow local api access when using Ruuvi Cloud
In 1.7.1 it is not possible to set authentication method when using Ruuvi Cloud. It would be nice to be able access the history api without sacrificing Ruuvi Cloud.

TheSomeMan:


ojousima (approved):
Thanks

PR #350, opened 2021-09-04
Fix #331: Password authentication fails when tried first time, pressing LOG IN again fixes the issue

### LINKED ISSUE #331: Password authentication fails when tried first time, pressing LOG IN again fixes the issue
When I press LOG IN button after typed my password, I get an error. Now, after I press LOG IN again, it works. Note that in this test my user name and password were both only 1 character long, but I tried also with longer user name and password, behaviour is the same.

image

TheSomeMan:


ojousima (approved):
Thanks

PR #354, opened 2021-09-09
Fix #353: UI: Access settings from LAN are reset after reconfiguring if they have not been changed

### LINKED ISSUE #353: UI: Access settings from LAN are reset after reconfiguring if they have not been changed

TheSomeMan:


ojousima (approved):
Thanks

PR #355, opened 2021-09-09
Fix #317: Configuring GW from the local network disconnects from the current (local) network

### LINKED ISSUE #317: Configuring GW from the local network disconnects from the current (local) network
When setting up the gateway via hotspot successfully (custom server, with lan access, ethernet DHCP) and then connect to gateway IP, the Configuration wizard is shown again.
If you try to configure it fails when setting ethernet DHCP and the gateway is no longer accessible from ethernet.

TheSomeMan:


ojousima (approved):
Thanks

PR #358, opened 2021-09-14
Close #352: Allow access from LAN by using the unique device ID as default password

### LINKED ISSUE #352: Allow access from LAN by using the unique device ID as default password
I think we don't have an issue for this. So I'm creating one.

TheSomeMan:


ojousima (approved):
Thanks

PR #359, opened 2021-09-17
Close #338: Adjust watchdog feeding rules

### LINKED ISSUE #338: Adjust watchdog feeding rules
Like discussed with Otso earlier, watchdog should be probably kicked for example in situations where no data has been sent successfully to the cloud in one hour. Also critical issues like too low heap memory should kick it. Please describe here how it's currently fed and then we can specify new rules and implement them.

TheSomeMan:


ojousima (approved):
Thanks

PR #366, opened 2021-09-24
Fix #365: UI: Connection to WiFi with a saved password does not work if WiFi AP temporarily disappears from the list of available networks

### LINKED ISSUE #365: UI: Connection to WiFi with a saved password does not work if WiFi AP temporarily disappears from the list of available networks

TheSomeMan:


ojousima (approved):
Thanks

PR #368, opened 2021-09-25
Fix #367: UI: Cloud options: Switching from MQTT to "Ruuvi Cloud" does not change the mode

### LINKED ISSUE #367: UI: Cloud options: Switching from MQTT to "Ruuvi Cloud" does not change the mode

TheSomeMan:


ojousima (approved):
Thanks

PR #371, opened 2021-09-27
Fix #370: MQTT does not start automatically after configuration, it starts after rebooting

### LINKED ISSUE #370: MQTT does not start automatically after configuration, it starts after rebooting

TheSomeMan:


main/mqtt.c
```diff
@@ -106,8 +133,43 @@ mqtt_publish_connect(void)
...
+static void
+mqtt_publish_state_offline(esp_mqtt_client_handle_t p_mqtt_client)
+{
+    char *p_message = "{\"state\": \"offline\"}";
+
+    mqtt_topic_buf_t topic;
+    mqtt_create_full_topic(&topic, g_gateway_config.mqtt.mqtt_prefix, "gw_status");
+    LOG_INFO("esp_mqtt_client_publish: topic:'%s', message:'%s'", topic.buf, p_message);
+    const int32_t mqtt_qos         = 0;
+    const int32_t mqtt_flag_retain = 0;
```
ojousima, line 164: The status message for online/offline could have retain_flag = 1, it would make sense to me that new client connecting would get retained gateway statuses.

ojousima (approved):
Let's discuss the retain flag before merge

ojousima:
Slack discussion: https://ruuvi.slack.com/archives/CFCSV42V8/p1632738188030900

main/mqtt.c
TheSomeMan: status message "online" has retain flag, but I'm not sure if it's needed for "offline" message because it is sent only when new configuration is sent to gw. 
When the gateway is reconfigured, MQTT client is stopped before applying new configuration and gw_status="offline" is sent at this moment.

ojousima: It seems to me that there's 3 possible states for MQTT:

1) Gateway is online and configured to use MQTT
2) Gateway is working and connected to Internet, but configured to not use MQTT / connect to different broker
3) Gateway is offline due to power / connectivity loss, LWT message state. 

It would make sense to have all these states in one topic and retained, there should not be any significant cost for the broker to retain one topic per gateway.

TheSomeMan: ok

PR #372, opened 2021-09-27
Close #242: Implement support for data transfer to the MQTT server without delay

### LINKED ISSUE #242: Implement support for data transfer to the MQTT server without delay

TheSomeMan:


ojousima (approved):
Thanks

PR #373, opened 2021-09-27
Close #287: UI: HTTP/MQTT: Replace the target radio buttons with checkboxes

### LINKED ISSUE #287: UI: HTTP/MQTT: Replace the target radio buttons with checkboxes
In addition to HTTP / MQTT mode, add an usage mode where GW is pollable only. This is for Home Assistant users who don't want to push their data to external servers

TheSomeMan:


ojousima (approved):
Thanks

PR #379, opened 2021-10-11
Close #374: Store the configuration in NVS in json format

### LINKED ISSUE #374: Store the configuration in NVS in json format
Currently the configuration in NVS is stored in BLOB.
When new fields is added or removed, other fields are shifted, in this case the configuration is cleared after installing the new fw and users need to reconfigure gateways which is not very convenient (and it might be a big problem if firmware is automatically updated).
Storing the configuration in json format will allow not to lose the current configuration after format changes.

TheSomeMan:


ojousima (approved):
Thanks

PR #390, opened 2021-10-17
Close #386: MQTT: Username and password need to support long values (128 for username and 256 for password)

### LINKED ISSUE #386: MQTT: Username and password need to support long values
For example Azure IoT Hub has autogenerated, long values for username and password: https://f.ruuvi.com/t/ruuvi-gateway-and-azure-iot-hub-connection-with-mqtt/5093/3?u=otso

```
According to Microsoft documentation the username must be something like “contoso.azure-devices.net/MyDevice01/?api-version=2018-06-30” and the password “SharedAccessSignature sig={signature-string}&se={expiry}&sr={URL-encoded-resourceURI}”. 
```

TheSomeMan:


ojousima (approved):
Thanks

PR #407, opened 2021-10-31
Fix #375: Wi-Fi networks with multiple hotspots aren't shown as one

### LINKED ISSUE #375: Wi-Fi networks with multiple hotspots aren't shown as one
Wi-Fi configurator doesn't show two hotspots as one and those networks are jumping on the configurator like shown on the video. Note also that the other one doesn't allow to enter password at all. Mobile phones and computers show only one network. 1.8.2 firmware. Video should explain the issue:

https://user-images.githubusercontent.com/1311912/136656477-e3c4cd0b-54dc-4a9e-a2bc-af94ddbcf9c2.mp4

TheSomeMan:


ojousima (approved):
Thanks

PR #408, opened 2021-11-01
Fix #406: MQTT/HTTP passwords are not saved

### LINKED ISSUE #406: MQTT/HTTP passwords are not saved
Description:

User configures MQTT server via configurator using local ethernet connection and adds long username and long password.
Actual: When user reopens the configurator he sees no asterisk in password field, assuming long password was not saved.
(for testing purposes I used a windows reg key)

TheSomeMan:


main/gw_cfg.c
```diff
@@ -95,13 +95,16 @@ gw_cfg_print_to_log(const ruuvi_gateway_config_t *const p_config)
...
+    LOG_DBG("config: mqtt password: %s", p_config->mqtt.mqtt_pass.buf);
     LOG_INFO("config: use http: %d", p_config->http.use_http);
     LOG_INFO("config: http url: %s", p_config->http.http_url.buf);
     LOG_INFO("config: http user: %s", p_config->http.http_user.buf);
     LOG_INFO("config: http pass: %s", "********");
+    LOG_DBG("config: http pass: %s", p_config->http.http_pass.buf);
     LOG_INFO("config: use http_stat: %d", p_config->http_stat.use_http_stat);
     LOG_INFO("config: http_stat url: %s", p_config->http_stat.http_stat_url.buf);
     LOG_INFO("config: http_stat user: %s", p_config->http_stat.http_stat_user.buf);
+    LOG_DBG("config: http_stat pass: %s", p_config->http_stat.http_stat_pass.buf);
```
ojousima, line 107: This seems a bit dangerous to me, if log level is changed to debug passwords will be printed in plaintext.

On the other hand, only way to print logs is to change firmware of the gateway and at that point attacker can just write the firmware to print out the passwords anyway. 

Please leave a comment at line 14, something like "// Warning: Debug log level prints out the passwords as a plaintext" so accidents won't happen.

ojousima (changes requested):
Thanks, please add the comment at line 14

ojousima (approved):
Thanks

PR #410, opened 2021-11-01
Close #409: Sort WiFi APs by RSSI to prevent loss of APs from higher channels when more than 16 APs are available

### LINKED ISSUE #409: Sort WiFi APs by RSSI to prevent loss of APs from higher channels when more than 16 APs are available
[link](https://github.com/ruuvi/esp32-wifi-manager/pull/130)

TheSomeMan:


ojousima (approved):
Thanks

PR #412, opened 2021-11-01
Fix #411: WiFi password is printed to log as a plaintext

### LINKED ISSUE #411: WiFi password is printed to log as a plaintext

TheSomeMan:


ojousima (approved):
Thanks

PR #414, opened 2021-11-01
Fix #413: Empty WiFi access points are shown if less than 16 access points are available

### LINKED ISSUE #413: Empty WiFi access points are shown if less than 16 access points are available

TheSomeMan:


ojousima (approved):
Thanks

PR #415, opened 2021-11-02
Fix #405: MQTT disconnection takes long time and causes user unable to access configurator

### LINKED ISSUE #405: MQTT disconnection takes long time and causes user unable to access configurator via local ethernet
Description:

FW is v1.8.3-25

User configures MQTT server with address mqtt.ruuvi.com and saves configuration.
Then he tries to go back to configurator via browser address 192.168.1.2 but it doesn't open.
Log is not showing any activity.
Then after a prolonged period of time address opens and user can access configurator again.

TheSomeMan:


ojousima (approved):
Thanks

PR #417, opened 2021-11-02
Fix #416: HTTP/MQTT passwords are printed to log as a plaintext

### LINKED ISSUE #416: HTTP/MQTT passwords are printed to log as a plaintext

TheSomeMan:


ojousima (approved):
Thanks

PR #419, opened 2021-11-02
Fix #418: The configuration is reset to the default value if the configuration wizard is closed before configuration is complete and Ethernet cable is connected

### LINKED ISSUE #418: The configuration is reset to the default value if the configuration wizard is closed before configuration is complete and Ethernet cable is connected

TheSomeMan:


ojousima (approved):
Thanks

PR #421, opened 2021-11-02
Fix #420: When LAN access is denied from everyone, UI is unclear

### LINKED ISSUE #420: When LAN access is denied from everyone, UI is unclear
[link](https://github.com/ruuvi/ruuvi.gwui.html/issues/143)

TheSomeMan:


ojousima (approved):
Thanks

PR #429, opened 2021-11-05
Fix #426: The gateway does not reconnect to WiFi after disconnecting from the hotspot

### LINKED ISSUE #426: The gateway does not reconnect to WiFi after disconnecting from the hotspot

TheSomeMan:


ojousima (approved):
Thanks

PR #430, opened 2021-11-05
Fix #427: WiFi configuration is reset after a manual firmware update

### LINKED ISSUE #427: WiFi configuration is reset after a manual firmware update

TheSomeMan:


ojousima (approved):
Thanks

PR #431, opened 2021-11-05
Fix #428: WiFi hotspot is automatically started after a manual firmware update from LAN

### LINKED ISSUE #428: WiFi hotspot is automatically started after a manual firmware update from LAN

TheSomeMan:


ojousima (approved):
Thanks

PR #434, opened 2021-11-14
Fix #422: User saves http server config with non-empty username and it shows Ruuvi cloud selected next time

### LINKED ISSUE #422: User saves http server config with non-empty username and it shows Ruuvi cloud selected next time
Description:

User had configured gateway to use MQTT server previously. He goes to configure settings again and now chooses http server and unselects MQTT, finishes configuration

Then he opens configurator again
Actual: Ruuvi cloud is now selected

TheSomeMan:


ojousima (approved):
Thanks

PR #436, opened 2021-11-17
Close #435: UI: Add a clarification where the user can get the default password for accessing from LAN

### LINKED ISSUE #435: UI: Add a clarification where the user can get the default password for accessing from LAN

TheSomeMan:


ojousima (commented):
Thanks, let's wait for the security group comment before sending the default auth flag

ojousima (approved):
Ok for now, has a risk of needing a change in future

ojousima (approved):
Thanks

PR #438, opened 2021-11-17
Fix #437: Firmware v1.9.1 is automatically downgraded to v1.8.2 after switching to "Manual update" mode if the gateway has not been rebooted

### LINKED ISSUE #437: Firmware v1.9.1 is automatically downgraded to v1.8.2 after switching to "Manual update" mode if the gateway has not been rebooted

TheSomeMan:


ojousima (approved):
Thanks

PR #446, opened 2021-12-02
Fix #441: If connecting to a Wi-Fi network fails, user is suggested to try again but it's not possible to try again

### LINKED ISSUE #441: If connecting to a Wi-Fi network fails, user is suggested to try again but it's not possible to try again
Sometimes connecting to a network fails like this:
Screenshot 2021-11-24 at 20 50 03

But it's not possible to try again by pressing the NEXT button because it's dim and not accessible. At this point, the only option is to press Back.
Screenshot 2021-11-24 at 20 51 41

(Build 461)

TheSomeMan:


ojousima (approved):
Thanks

PR #447, opened 2021-12-08
#441: Fix: UI: Input password block is displayed for WiFi without auth at the first time after scanning WiFi

TheSomeMan:


ojousima (approved):
Thanks

Issue #449, opened 2021-12-30
Custom http server seems to only support https


Scrin:
It took me way too long to figure it out, but it seems that only https is supported in custom http server. Not sure if I'm doing something wrong, but it seems that if I configure anything but a url starting with `https://`, the gateway doesn't send anything to the configured custom http server. Support for plain http would be useful for very local or "offline" setups where dealing with https certificates is unnecessary.

ojousima:
This is related to ESP SDK supporting asynchronous http posts only when using https, https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_client.html#_CPPv4N24esp_http_client_config_t8is_asyncE. Using asynchronous calls solves crashes when server replies slowly, but we noticed afterwards that HTTP is broken. 

Fix is in the works and will be included in the next release.

yob:
Is this a regression?

I've been using a custom HTTP endpoint with my ruuvi gateway for a few months without issue, but after upgrading from 1.8.2 to 1.9.2 I've just been hit by this issue.

ojousima:
Yes, this is a regression. 

> As a workaround, will the gateway post to a HTTPS server using a self signed certificate?

@TheSomeMan ^?

It's also possible to update back to v1.8.2 and select manual updates only until the issue is fixed

TheSomeMan:
> As a workaround, will the gateway post to a HTTPS server using a self signed certificate?

this is not supported at the moment, requests to the server are not signed, but it can be implemented in the future.

> It's also possible to update back to v1.8.2 and select manual updates only until the issue is fixed

yes, of course, it's possible to downgrade to v1.8.2

TheSomeMan:
Fixed in #453 and #452

PR #453, opened 2022-01-01
#449: Patch esp_http_client and tcp_transport to implement asynchronous mode support for HTTP connections

TheSomeMan:


ojousima (approved):
Thanks, please update the patch notes too

PR #457, opened 2022-01-13
Fix #452: Gateway enters a boot loop if HTTP POST times out 

### LINKED ISSUE #452: Gateway enters a boot loop if HTTP POST times out
If a user has configured to use HTTP with their own or 3rd party server and the said server goes down such that all requests time out, the gateway enters a boot loop, rebooting approximately every 10-20 seconds. This does not happen if the request simply fails, so it seems to be related to timeouts only.

This is extra problematic as you can't finish the configuration wizard to remove the http configuration before the gateway reboots. I ran into this issue after enabling the http post feature for testing, and then after leaving my test server off, I found that the gateway entered a reboot loop, uptime never exceeding 20 seconds.

Here are the metrics from the /metrics endpoint, you can see that the gateway was in a reboot loop from 2:30 to 04:00 when my test server was down, and another time around 12:40 when I took the test server down again. Probably unrelated but it also looks like the free heap was acting weird during the boot loop phase.

image

TheSomeMan:


ojousima (approved):
Thanks, great attention on detail with change log. Hopefully this will also fix the reboots we observe in statistics

PR #459, opened 2022-01-13
Fix #458: Sending http_stat does not get stored to flash if http data send is disabled

### LINKED ISSUE #458: v1.9.2: Sending http_stat does not get stored to flash if http data send is disabled.
If user configures the Gateway to send data with MQTT but not to a HTTP endpoint and selects http_stat to send diagnostic data to Ruuvi, http_stat setting stays in effect only until next reboot. 

If user selects HTTP to Ruuvi Cloud, custom MQTT and http_statistics, statistics get sent.

Expected behavior is that HTTP statistics would be sent independently of HTTP data.

TheSomeMan:


ojousima (approved):
Thanks

PR #462, opened 2022-01-14
Close #450: Add info metric to the /metrics endpoint

### LINKED ISSUE #450: Info metric to the /metrics endpoint
It would be useful to have an info metric added to the /metrics endpoint (metrics.c) in the "standard" info metric format for rarely changing diagnostic info, something like:

```
ruuvigw_info{mac="FF:EE:DD:CC:BB:AA",esp_fw="v1.9.2",nrf_fw="v0.7.2"} 1
```

where the mac is the mac of the gateway, and the fw versions are the current firmware versions. I'm mainly interested in having the firmware versions available for tracking with standard prometheus-compatible monitoring solutions. The info metric could also include other rarely changing attributes such as other build infos, if available.

This should be relatively trivial to add, though I don't have a usable development setup at the moment

TheSomeMan:


ojousima (approved):
Thanks. Let's merge this earliest on Monday morning 17.1, that way we don't need to create extra commit just to reset the 2-week timer for release rollout and we can include it in v1.10.0 release

PR #469, opened 2022-01-17
Fix #467: Race condition with MQTT task causes network watchdog false triggering and gateway reboot

### LINKED ISSUE #467: Race condition with MQTT task causes network watchdog false triggering and gateway reboot
Description:

Looking at logs I see a reboot rst:0xc (SW_CPU_RESET),boot:0x17 (SPI_FAST_FLASH_BOOT)
[loki.txt](https://github.com/ruuvi/ruuvi.gateway_esp.c/files/7878846/loki.txt)

TheSomeMan:


ojousima (approved):
Thanks

PR #478, opened 2022-02-09
Close #456: Update configuration mechanism - Encrypt hotspot data

### LINKED ISSUE #456: Update configuration mechanism - Encrypt hotspot data
Current configuration mechanism has security and usability issues:

* Attacker can record all WiFi traffic in the area and crack user configuration (which includes user home WiFi) if credentials are entered over WiFi hotspot
* Updating large installations is cumbersome because each Gateway has to be configured manually.

To fix these issues, following things should be done:

* Allow Gateway to fetch a configuration .json file from a preconfigured server. By default this is empty field, add URL to fetch configuration from to under Gateway Update page.
* Encrypt configuration posted to Gateway HTTP Server using asymmetric cryptography. Details such as is RSA or Diffie-Hellman used and which exact libraries are used are to be discussed.
* Add CRC32 of configuration to Gateway statistics and under /metrics endpoint. 

See also #455

TheSomeMan:


ojousima (approved):
Thanks, let's merge the MQTT fix first

ojousima:
MQTT is ok, ok to merge

PR #483, opened 2022-02-14
Fix #482: LAN_AUTH_APIKEY is printed as plaintext to debug console

### LINKED ISSUE #482: v1.11.0: LAN_AUTH_APIKEY is printed as plaintext to debug console
Should be "***" instead

TheSomeMan:


ojousima (approved):
Thanks

TheSomeMan:
To test it, you need to configure bearer authentication on the Access Settings page and check that the bearer key is never printed to the log console.

ojousima:
Tested on v1.12.1

PR #487, opened 2022-02-17
Close #484: Use open security for WiFi access point

### LINKED ISSUE #484: Use open security for WiFi access point
As the data is encrypted in application layer, Gateway hotspot does not need a password anymore

TheSomeMan:


ojousima (approved):
Thanks

PR #489, opened 2022-02-18
Fix #488: RSSI drop after update from v1.9.2-> newer

### LINKED ISSUE #488: RSSI drop after update from v1.9.2-> newer
Samples:
image
image
image

This is graph of reverting v1.11.0 to v1.9.2:
image

LNA GPIO control and WiFi coexistence needs to be checked.

TheSomeMan:


ojousima (approved):
Thanks

PR #492, opened 2022-02-23
#488: Disable the use of GPIO LNA_CRX until this bug is fixed: https://github.com/espressif/esp-idf/issues/6930

TheSomeMan:


ojousima (approved):
Thanks

PR #498, opened 2022-03-23
Close #454: Create a configuration file

### LINKED ISSUE #454: Create a configuration file
Some of our customers would greatly benefit of pre-configured Ruuvi Gateways. To be able to configure the devices during production, configuration registers should be possible to be written when flashing the ESP32. The most simple solution would be probably to have an additional file to be flashed:

- bootloader.bin
- fatfs_gwui.bin
- fatfs_nrf52.bin
- ota_data_initial.bin
- partition-table.bin
- ruuvi_gateway_esp.bin
- **config.bin (NEW FILE)**

We could offer also a public web page that allows users to easily generate desired configs.

TheSomeMan:


TheSomeMan:
> Thanks, this also needs an update to readme for flashing if I understand correctly?

Yes, you're right, it slipped my mind, thanks

ojousima (approved):
Thanks

DG12:
Curious as to why this changed from a configuration file to a partition.
Was it so that both nRF and ESP could access it?

TheSomeMan:
> Curious as to why this changed from a configuration file to a partition. Was it so that both nRF and ESP could access it?

If I understand you question correctly, no, only ESP32 can access the partition with the configuration file.
And we didn't have a configuration file before, the default settings were per-compiled to the binary firmware image.

DG12:
Why then a partition and not just a file?

ojousima:
Partition allows for easy creation and flashing of default configuration, without having to recompile the entire image or build a mechanism for fetching the file from somewhere

PR #502, opened 2022-04-22
Close #455: Request new configuration remotely

### LINKED ISSUE #455: Request new configuration remotely
During the configuration wizard, one should be able to define a remote URL from where the Gateway would poll new configuration every now and then.

This is related to the issue https://github.com/ruuvi/ruuvi.gateway_esp.c/issues/454

The new config could be either a bin file or a json. If we'll have an easy to use config creation web page that outputs .bin as suggested in the issue https://github.com/ruuvi/ruuvi.gateway_esp.c/issues/454, maybe that would be the way to go.

TheSomeMan:


ojousima (approved):
Thanks

PR #504, opened 2022-04-23
Fix #503: MQTT over SSL does not work after downgrading esp-idf from v4.3.3 to v4.2.2

### LINKED ISSUE #503: MQTT over SSL does not work after downgrading esp-idf from v4.3.3 to v4.2.2

TheSomeMan:


ojousima (approved):
Thanks

PR #506, opened 2022-04-24
Fix #505: Deadlock in mqtt after saving configuration causes task watchdog triggering and gateway rebooting

### LINKED ISSUE #505: Deadlock in mqtt after saving configuration causes task watchdog triggering and gateway rebooting

TheSomeMan:


ojousima (approved):
Thanks

PR #518, opened 2022-05-08
Fix #517: After setting up gateway via WiFi hotspot, it does not reconnect to WiFi access point after losing connection

### LINKED ISSUE #517: After setting up gateway via WiFi hotspot, it does not reconnect to WiFi access point after losing connection
If WiFi hotspot on the gateway was activated and the gateway was configured through Web UI, the automatic WiFi reconnection stops working until the gateway is rebooted. If WiFi connection is lost, the gateway will be restarted after 1 hour by a network watchdog.

TheSomeMan:


ojousima (approved):
Thanks

PR #522, opened 2022-05-17
Fix #521: Statistics is always sent to the remote server even if it is disabled in the configuration

### LINKED ISSUE #521: Statistics is always sent to the remote server even if it is disabled in the configuration

TheSomeMan:


ojousima (changes requested):
Please add a unit test to for this configuration

ojousima (approved):
Let's proceed without unit test change after all

PR #530, opened 2022-05-19
Fix #529: v1.12.0: Unable to update firmware due to download timeout

### LINKED ISSUE #529: v1.12.0: Unable to update firmware due to download timeout

TheSomeMan:


ojousima (approved):
Thanks

PR #537, opened 2022-06-12
Close #536: By default, the network connection type should be Ethernet

### LINKED ISSUE #536: By default, the network connection type should be Ethernet

TheSomeMan:


ojousima (approved):
Thanks

PR #541, opened 2022-06-18
Fix #538: On a slow network connection, the file download may stop, making it impossible to download the software update

### LINKED ISSUE #538: On a slow network connection, the file download may stop, making it impossible to download the software update

TheSomeMan:


ojousima (changes requested):
Please rather comment out the `read_len = 0` and add comment explaining the change, e.g.

```
 // Patch: Remove assigning read_len to 0 to avoid downloads stopping on slow connection
 // read_len = 0;
```

TheSomeMan:
> Please rather comment out the `read_len = 0` and add comment explaining the change, e.g.
> 
> ```
>  // Patch: Remove assigning read_len to 0 to avoid downloads stopping on slow connection
>  // read_len = 0;
> ```

I'm not sure if it's necessary, because it's a bug in my previous patch 0798fe2ecdbb48be94c1b64323a24e15152d3c7a (#452), 
see the difference:
image

ojousima (approved):
As the len=0 change was from previous patch, no comment needed

PR #545, opened 2022-06-21
Fix #534: The "/history" endpoint does not work correctly when time synchronization is disabled

### LINKED ISSUE #534: The "/history" endpoint does not work correctly when time synchronization is disabled

TheSomeMan:


ojousima (approved):
Thanks

PR #556, opened 2022-06-29
Fix #555: HTTP redirection for downloaded files does not work if server return HTTP error 301 (Moved Permanently)

### LINKED ISSUE #555: HTTP redirection for downloaded files does not work if server return HTTP error 301 (Moved Permanantly)

TheSomeMan:


ojousima (approved):
Thanks

Issue #535, opened 2022-06-12
UI improvements


TheSomeMan:

markoaamunkajo:
All child cards QA passed, calling this Done

PR #559, opened 2022-07-02
#535: UI: Add word-wrapping for buttons to fix the problem with "Continue without updating" button

TheSomeMan:


ojousima (approved):
Thanks

PR #572, opened 2022-07-07
Fix #571: Sometimes task watchdog got triggered when Ethernet is started

### LINKED ISSUE #571: Sometimes task watchdog got triggered when Ethernet is started

TheSomeMan:


ojousima:
Is there something that will detect if `esp_err_t err = esp_eth_start(g_eth_handle);` gets stuck?

TheSomeMan:
> Is there something that will detect if `esp_err_t err = esp_eth_start(g_eth_handle);` gets stuck?

yes, you're right, there is a better solution, see my last commit

ojousima (approved):
Thanks

PR #586, opened 2022-07-12
Fix #585: Incorrect redirect to 10.10.0.1 when accessing from LAN by dynamic DNS name

### LINKED ISSUE #585: Incorrect redirect to 10.10.0.1 when accessing from LAN by dynamic DNS name

TheSomeMan:


ojousima (approved):
Thanks

PR #588, opened 2022-07-13
Fix #581: MQTT continues to work even when WiFi hotspot is activated

### LINKED ISSUE #581: MQTT continues to work even when WiFi hotspot is activated

TheSomeMan:


ojousima (approved):
Thanks

PR #589, opened 2022-07-13
Fix #582: wifi_manager can consume a lot of CPU if it cannot connect to WiFi

### LINKED ISSUE #582: wifi_manager can consume a lot of CPU if it cannot connect to WiFi

TheSomeMan:


ojousima (approved):
Thanks

PR #592, opened 2022-07-21
Fix #590: Log gets filled with no network connection messages

### LINKED ISSUE #590: Log gets filled with no network connection messages
Description:

User configured test.mosquitto.net with wrong settings to Ruuvi configurator and saved.
Now logs are full of entries for no network connection

Solution:
Print once per minute, just to see that MQTT is being used

TheSomeMan:


ojousima (approved):
Thanks

PR #594, opened 2022-07-25
Fix #593: MQTT: use MAC-address as Client ID if client_id is not set in configuration

### LINKED ISSUE #593: MQTT: use MAC-address as Client ID if client_id is not set in configuration
When requesting configuration from the remote server and ClientID is not set, then MQTT cannot be initialized:
```
I (559045) MQTT: [mqtt_task/5] MQTT_EVENT_BEFORE_CONNECT
E (559094) MQTT_CLIENT: Connect message cannot be created
E (559095) MQTT_CLIENT: MQTT connect failed
I (559096) MQTT: [mqtt_task/5] MQTT_EVENT_DISCONNECTED
```

TheSomeMan:


ojousima (approved):
Thanks

PR #604, opened 2022-08-09
Close #602: Turn off DNS server as soon as it is no longer needed

### LINKED ISSUE #602: Turn off DNS server as soon as it is no longer needed

TheSomeMan:


ojousima (approved):
Thanks

PR #606, opened 2022-08-09
Fix #605: Due to the race condition, sometimes HTTP response without a body is returned in response to POST /connect.json

### LINKED ISSUE #605: Due to the race condition, sometimes HTTP response without a body is returned in response to POST /connect.json
[link](https://github.com/ruuvi/esp32-wifi-manager/issues/195)

TheSomeMan:


ojousima (approved):
Thanks

Issue #601, opened 2022-08-08
WiFi hotspot disconnects client after it tries to connect the gateway to WiFi network


TheSomeMan:
```
I (287957) http_server: [http_server/4] Request from 10.10.0.2 to 10.10.0.1 (Host: 10.10.0.1): POST /connect.json
I (287958) http_server: [http_server/4] POST /connect.json
I (287964) http_server: [http_server/4] http_server_netconn_serve: POST /connect.json
I (287974) http_server: [http_server/4] POST /connect.json: SSID:WIFI768_2.4G, PWD: NULL - reconnect to saved WiFi
I (287983) http_server: [http_server/4] Json resp: code=200, content:
{}
I (287991) http_server: [http_server/4] Response: OK
I (288091) http_server: [http_server/4] Request from 10.10.0.2 to 10.10.0.1 (Host: 10.10.0.1): GET /status.json
I (288093) http_server: [http_server/4] Json resp: code=200, content:
{"ssid":"WIFI768_2.4G","ip":"0","netmask":"0","gw":"0","dhcp":"","urc":2,"lan":0}

I (288106) http_server: [http_server/4] Response: OK
I (288983) wifi_manager: [Tmr Svc/24] wifi_manager_timer_cb_reconnect: wifiman_msg_send_cmd_connect_sta: CONNECTION_REQUEST_AUTO_RECONNECT
I (288985) wifi_manager: [Tmr Svc/24] Send msg: ORDER_CONNECT_STA, conn_req_code=2
I (289009) wifi_manager: [wifi_manager/5] MESSAGE: ORDER_CONNECT_STA: CONNECTION_REQUEST_AUTO_RECONNECT, event_bits=0x041d
I (289020) wifi_manager: [wifi_manager/5] WIFI_MANAGER:EV_STATE: Clear WIFI_MANAGER_REQUEST_DISCONNECT_BIT
I (290641) wifi:switch to channel 3
I (290642) wifi:ap channel adjust o:1,0 n:3,0
I (290642) wifi:new:<3,0>, old:<1,0>, ap:<3,0>, sta:<3,0>, prof:1
I (290645) wifi:new:<3,0>, old:<3,0>, ap:<3,0>, sta:<3,0>, prof:1
I (290650) wifi:state: init -> auth (b0)
I (290678) wifi:state: auth -> assoc (0)
I (290690) wifi:state: assoc -> run (10)
[...]

PR #612, opened 2022-08-11
Close #603, #601: Save preferred WiFi channel for AP when client tries to connect to WiFi network

### LINKED ISSUE #603: Save preferred WiFi channel for AP when client tries to connect to WiFi

TheSomeMan:


ojousima:
Does this affect remote configuration JSON too?

TheSomeMan:
> Does this affect remote configuration JSON too?

One new field "wifi_ap_config/channel" was added to JSON, but if it does not exist in JSON it's okay, default value 1 will be used.
So, if new firmware downloads old JSON, then it will use the default value 1. If old firmware downloads new JSON, then it will just ignore "wifi_ap_config/channel" field and use the hardcoded value 1.

ojousima (approved):
thanks

PR #617, opened 2022-08-24
Fix #614: WiFi hotspot does not start after erasing all flash memory

### LINKED ISSUE #614: WiFi hotspot does not start after erasing all flash memory

TheSomeMan:


main/http_server_cb.c
```diff
@@ -902,6 +902,11 @@ http_server_cb_on_post_ruuvi(const char *p_body)
         {
             ethernet_update_ip();
         }
+        const force_start_wifi_hotspot_t force_start_wifi_hotspot = settings_read_flag_force_start_wifi_hotspot();
+        if (FORCE_START_WIFI_HOTSPOT_PERMANENT == force_start_wifi_hotspot)
+        {
+            settings_write_flag_force_start_wifi_hotspot(FORCE_START_WIFI_HOTSPOT_DISABLED);
```
ojousima, line 908: Is it intentional to write `FORCE_START_WIFI_HOTSPOT_DISABLED` to settings inside 
```
if (FORCE_START_WIFI_HOTSPOT_PERMANENT == force_start_wifi_hotspot)
{
}
```
?

ojousima (approved):
Thanks, please check comment

main/http_server_cb.c
TheSomeMan: I added a comment:

```
            /* A permanent start-up of the Wi-Fi hotspot should be performed after each reboot
             * only until a new non-default configuration is saved */

```

PR #618, opened 2022-08-24
Fix #615: Turn on solid green LED only after Ruuvi cloud has returned OK message

### LINKED ISSUE #615: Turn on solid green LED only after Ruuvi cloud has returned OK message

TheSomeMan:


ojousima (approved):
Thanks

PR #619, opened 2022-08-25
Fix #289: Gateway CSS doesn't always load

### LINKED ISSUE #289: Gateway CSS doesn't always load
This is possibly related to #253, image below. This was the view when browsing to 10.10.0.1, fixed in refresh
image

TheSomeMan:


ojousima (approved):
Thanks

PR #620, opened 2022-08-25
#614: Disable Wi-Fi hotspot activation after applying auto-configuration by connecting Ethernet cable

TheSomeMan:


ojousima (approved):
Thanks, log explanations do a good work of clarifying the program intent

ojousima (approved):
Thanks

PR #622, opened 2022-08-29
Close #621: Check if the configuration button is pressed at the earliest stage after rebooting

### LINKED ISSUE #621: Check if the configuration button is pressed at the earliest stage after rebooting
The state of the button is currently checked 3 seconds after a reboot, and if the user accidentally presses the button at this point, the gateway configuration will be erased. 
In addition, the user must press and hold the button for at least 5+3=8 seconds to clear the gateway configuration.

TheSomeMan:


ojousima (approved):
Thanks

PR #627, opened 2022-08-31
#614: Force start Wi-Fi hotspot when Ethernet is not connected even if gateway has non-default configuration

TheSomeMan:


ojousima (approved):
Thanks

PR #636, opened 2022-09-13
Fix #635: Configuration is erased after a rollback from v1.12.2 to v1.11.2

### LINKED ISSUE #635: Configuration is erased after a rollback from v1.12.2 to v1.11.2

TheSomeMan:


ojousima (approved):
Thanks

PR #641, opened 2022-09-21
Close #638: Put data in the history buffer, even if the time is not synchronized yet

### LINKED ISSUE #638: Put data in the history buffer, even if the time is not synchronized yet

TheSomeMan:


ojousima (approved):
Thanks

PR #645, opened 2022-10-07
Close #630: UI: Disable the use of API-key on Access settings page when "Remote configurable without a password" option is set

### LINKED ISSUE #630: Confusing behavior when supplying a Bearer token without enabling authentication
It seems that if one does not enable authentication on the gateway and then tries to poll the /history endpoint for example with any Bearer authentication, the gateway rejects the call with http error 401. I would expect the gateway to completely ignore the Authorization header if authentication is not enabled.

Steps to reproduce:
- Configure a gateway with no authentication (use the "Remote configurable without a password (unsafe, not recommended)" option in Access Settings page)
- Try to call the /history endpoint with `curl -v gateway.ip.address/history` and notice it's working as expected
- Try to call the /history endpoint with `curl -vH"Authorization: Bearer foo" gateway.ip.address/history` and notice the gateway returning a http 401 error back, even though authentication should be disabled

TheSomeMan:


ojousima (approved):
Thanks

PR #647, opened 2022-10-12
Add script build_release.sh and script/log_remove_timestamps.py

TheSomeMan:


DG12:
echo -e "\033[31mCurrent version is not a release version: $FW_TAG" 1>&2

which sets Bold  should end with [30m which Clears all special attributes.

TheSomeMan:
> echo -e "\033[31mCurrent version is not a release version: $FW_TAG" 1>&2
> 
> which sets Bold should end with [30m which Clears all special attributes.

Thanks

PR #649, opened 2022-10-14
Format source code using clang-format-14

TheSomeMan:


ojousima (approved):
Thanks

DG12:
Can you please explain what SonarCloud Code Analysis "3 more branches need to be covered by tests " means.

TheSomeMan:
> Can you please explain what SonarCloud Code Analysis "3 more branches need to be covered by tests " means.

This refers to unit testing. First of all, SonarCloud checks if every line of code is covered by tests. Next, it checks if every condition in the "if-clause" is covered by tests.
For example:
```
if (a && b) {
  // ...
}
```
Such "if-clause" should be covered by at least 4 unit tests:
* a == true and b == true
* a == true and b == false
* a == false and b == true
* a == false and b == false 

"3 more branches need to be covered by tests " means that we need to write at least 3 additional unit tests for if-clauses so that the coverage threshold (65%) is reached.

DG12:
Thank you very much for that explanation.

PR #653, opened 2022-10-19
Fix #544: Static Ethernet IP address fails unless all fields are given

### LINKED ISSUE #544: v1.12.1: Static Ethernet IP address fails unless all fields are given
Configuring only static IP address, Gateway address, subnet mask and leaving DNS addresses blank leads to Gateway failing DNS queries and rebooting.

Steps to repeat: Configure Static Ethernet IP address, leave DNS addresses blank. 

My expectation would be that blank DNS addresses would be fetched via DHCP. Marking this as an usability issue for v1.13.0

image

TheSomeMan:


ojousima (approved):
Thanks

PR #656, opened 2022-10-20
Close #655: UI: Do not refresh the page with the list of Wi-Fi if the list has not actually been changed

### LINKED ISSUE #655: UI: Do not refresh the page with the list of Wi-Fi if the list has not actually been changed
[link](https://github.com/ruuvi/ruuvi.gwui.html/issues/235)

TheSomeMan:


ojousima (approved):
Thanks

PR #657, opened 2022-10-20
Fix #648: Static IP configuration is lost after downloading an empty configuration from a remote server

### LINKED ISSUE #648: Static IP configuration is lost after downloading an empty configuration from a remote server
[link](https://f.ruuvi.com/t/configuration-with-json/5689/9)

TheSomeMan:


ojousima (approved):
Thanks

PR #664, opened 2022-11-15
Close #663: Do no allow to access from LAN while WiFi hotspot is active

### LINKED ISSUE #663: Do no allow to access from LAN while WiFi hotspot is active
[link](https://github.com/ruuvi/esp32-wifi-manager/issues/190)

TheSomeMan:


ojousima (approved):
Thanks

PR #674, opened 2022-12-07
Close #660: Send reset reason and other debug info in http statistics

### LINKED ISSUE #660: Send reset reason in http statistics

TheSomeMan:


ojousima (approved):
Thanks, impressive usage of linker to wrap the reset handlers.

PR #679, opened 2022-12-11
Fix #671: Network watchdog is triggered if there are no Ruuvi sensors nearby

### LINKED ISSUE #671: Network watchdog is triggered if there are no Ruuvi sensors nearby
The gateway does not try to send data to the cloud if it has not received any data from the sensors. Because of this, the network connection status is not updated and LED indication mode is wrong.

TheSomeMan:


ojousima (approved):
Thanks,

Is there something monitoring the nRF52 itself? nRF52 has a watchdog, so if it times out it would request settings on boot from Gateway, so nRF52 reboot will fix itself?

TheSomeMan:
> Thanks,
> 
> Is there something monitoring the nRF52 itself? nRF52 has a watchdog, so if it times out it would request settings on boot from Gateway, so nRF52 reboot will fix itself?

Right, if there are no Ruuvi sensors nearby, then nRF52 restarts every 10 (or 15?) seconds and requests settings.

PR #686, opened 2023-01-25
Close #92: Backend check / validation

### LINKED ISSUE #92: Feature: Backend check / validation
Warn user about malformed URL, leading / trailing slash in MQTT topic etc. 

Possibly establish connection to Internet before configuring the Backend and report if backend can be connected to?

TheSomeMan:


ojousima (approved):
Thanks, I'm amazed of the speed you write code :) 

Please send me an example of new statistic JSON message when you have the chance, I'll look into adding support for new fields into our telemetry.

PR #689, opened 2023-02-07
Close #237: Update README.md

### LINKED ISSUE #237: Update Readme
Rewrite the readme.md so that it'll have all the necessary information available to flash and read console log using different operating systems. Explain how ESP is flashing the nRF and how long it may take. Avoid instructing users to find help from 3rd party websites. Remove redundant pin definition informations that are referring to devkits that were used a long time ago. Older than modified (R15 disconnected) Ruuvi Gateway A2 boards don't need to be supported in the documentation.

TheSomeMan:


ojousima (approved):
Thanks, very detailed

PR #690, opened 2023-02-07
#237: Add Ruuvi Gateway configuration guide for advanced users (gateway_configuration.md)

TheSomeMan:


ojousima:
Moving review to @markoaamunkajo

markoaamunkajo (commented):
Should the document mention about footer with current fw version and nRF fw version?
If so, then would be good to maybe update the screenshots to include these and add a general topic to document.,

markoaamunkajo (approved):

markoaamunkajo (approved):

Issue #699, opened 2023-02-22
Enable mDNS service to resolve the Gateway hostname


TheSomeMan:

markoaamunkajo:
QA passed

PR #702, opened 2023-02-23
#699: Update mDNS component to v1.0.8 to fix the long response and timeout problem on Ubuntu

TheSomeMan:


ojousima:
Would it be possible to add mdns component as a git submodule?

TheSomeMan:
> Would it be possible to add mdns component as a git submodule?

Espressif moved some modules to an independent github repo: https://github.com/espressif/esp-protocols
mdns is located as a folder in https://github.com/espressif/esp-protocols/tree/master/components/mdns, so I don't think that it's possible to convert it into a standalone submodule.
I downloaded it as an archive from here: https://components.espressif.com/components/espressif/mdns

ojousima (approved):
Thanks

PR #711, opened 2023-03-08
Close #710: Change name of HTTP/Websocket client from "ESP HTTP Client" to "Ruuvi HTTP Client"

### LINKED ISSUE #710: Change name of HTTP/Websocket client from "ESP HTTP Client" to "Ruuvi HTTP Client"

TheSomeMan:


ojousima:
The name should be something that clearly identifies the device as a Ruuvi Gateway, right now we scan our backend logs for `CLIENT LIKE %ESP32%` to monitor incoming Gateway traffic. `Ruuvi` as a client could be any application. I'd suggest using something like "Ruuvi Gateway" or "RGW" for short.

ojousima (approved):
Thanks

PR #713, opened 2023-03-08
Fix #712: WiFi hotspot may erroneously start after a reboot due to race conditions when Ethernet is connected

### LINKED ISSUE #712: WiFi hotspot may erroneously start after a reboot due to race conditions when Ethernet is connected

TheSomeMan:


ojousima (approved):
Thanks. I think that the maximum time ethernet takes to start was well-defined, it was related to the time oscillator gets a stable clock after being enabled. @laurijamsa thoughts?

Issue #704, opened 2023-03-01
UI: Backend validation: show error message if validation fails


TheSomeMan:
[link](https://github.com/ruuvi/ruuvi.gwui.html/issues/276)

markoaamunkajo:
QA passed, discussed on channel

PR #714, opened 2023-03-11
#704: UI: Backend validation: fix the problem with json response parsing error if the response contains line feed

TheSomeMan:


ojousima (approved):
Thanks

PR #717, opened 2023-03-13
Close #715: Use bootloader and partition_table from v1.9.2

### LINKED ISSUE #715: Use bootloader and partition_table from v1.9.2

TheSomeMan:


ojousima (approved):
Thanks

PR #746, opened 2023-05-02
Close #695: Gateway should try to resend data at 1 minute interval after error

### LINKED ISSUE #695: Gateway should try to resend data at 1 minute interval after error
Currently Gateways send at 10 second interval, which causes a traffic spike on backend problem which can make the problem worse leading to an outage

TheSomeMan:


ojousima (approved):
Thanks

PR #753, opened 2023-05-05
Close #238: Implement blacklist/whitelist for sensor MAC addresses

### LINKED ISSUE #238: Implement blacklist/whitelist for sensor MAC addresses
In some cases it would be beneficial to being able to define which Ruuvi sensors (or other nearby Bluetooth broadcast transmitters) to forward via HTTP/MQTT. Some reasons:

- Data plan and/or bandwidth of Gateway's internet connection is limited and too many Ruuvi or other BLE devices around
- Gateway is pushing data to a 3rd party server and unnecessary incoming data is causing extra load and requires additional filtering
- A 3rd party server is used and use case requires Gateway to stay quiet by default and send any data only if a certain predefined BLE device(s) comes in the range of the Gateway

TheSomeMan:
User interface likely to be redesigned


ojousima (approved):
Thanks. Just to verify that I understood correctly: The Gateway collects all seen sensors to internal buffer, but allows only whitelisted sensors into data sent to backend?

ojousima:
Also, does this filter sensors sent via MQTT too?

TheSomeMan:
> Thanks. Just to verify that I understood correctly: The Gateway collects all seen sensors to internal buffer, but allows only whitelisted sensors into data sent to backend?

No, if MAC does not pass the filter, it will be ignored and not put into the internal buffer.

> Also, does this filter sensors sent via MQTT too?

This filter applies to all targets, so if MAC is filtered out, then it will not be put into the buffer for HTTP retransmission and will not be relayed to MQTT.

```
static void
adv_post_send_report(void* p_arg)
{
    // ...
    
    if (adv_post_check_if_mac_filtered_out(p_cfg_cache, &adv_report.tag_mac))
    {
        adv_post_cfg_access_mutex_unlock(&p_cfg_cache);
        return;
    }    
    
    // ...
}
```

ojousima:
> No, if MAC does not pass the filter, it will be ignored and not put into the internal buffer.

Does this mean that once whitelist is configured, no new tags can be added to whitelist without resetting the whitelist as new tags are not passed to UI from internal buffer?

TheSomeMan:
> > No, if MAC does not pass the filter, it will be ignored and not put into the internal buffer.
> 
> Does this mean that once whitelist is configured, no new tags can be added to whitelist without resetting the whitelist as new tags are not passed to UI from internal buffer?

Yes, that's true. I added a comment about it to UI:

image

In any case, we have another filter in nRF52 (by the company_id) and if the user wants to see all sensors they need to disable filtering by company_id, save config and return to the scan configuration page.

Theoretically, this can be implemented by maintaining a separate buffer of available MAC addresses and dynamically reconfiguring on nRF52 while Web UI is active.

TheSomeMan:
Also, in the most useful scenario, when the 'Blacklist' mode is active, the user will see filtered-out sensors in the 'Configured Filters' area.

ojousima:
Ok, with the note about list not showing new sensors through existing whitelist I think this is all good

ojousima (approved):
Thanks

Issue #597, opened 2022-07-30
Support 2 HTTP targets simultaneously


ojousima:
E.G.
```
Send data to Ruuvi Cloud   [X]
Send data to other service [X]
  > URL: https://example.iot-integration.com/ruuvi
  > Format: 
    > Ruuvi ( )
    > Influx ( )
    > . . .
    > Example IOT Format (X)
```

Scrin:
I think the format selection should happen before the URL selection, since some formats may require additional or different parameters, for example different authentication methods. Perhaps something like this:
```
Send data to Ruuvi Cloud   [X]
Send data to other service [X]
  > Format: 
    > Ruuvi ( )
    > Influx (X)
      > URL: https://example.com/my-influxdb-sits-in-a-subpath
      > org: tunkki
      > bucket: ruuvi
      > token: myauthtokenforinfluxdb
    > . . .
    > Example IOT Format ( )
```
which would then construct the "actual" influxdb url in the UI side like this (assuming the gateway would be using the v2 api and ms precision for the timestamps): `https://example.com/my-influxdb-sits-in-a-subpath/api/v2/write?org=tunkki&bucket=ruuvi&precision=ms` and use the token in the `Authorization: Token <token>` header.

Of course all that could be input in a single url field, but that's rather inconvenient for the user with some protocols. For example in the InfluxDB example above, the user would need to know that the gateway uses the v2 api and uses ms precision and add/set those appropriately to the url.

PR #754, opened 2023-05-07
#597: Fix: Disabling sending to Ruuvi target does not work, LED works incorrectly when only custom HTTP target is used

TheSomeMan:


ojousima (approved):
Thanks

PR #759, opened 2023-05-11
Fix #758: mqtt-client: It's not possible to use custom SSL certificate for a server

### LINKED ISSUE #758: mqtt-client: It's not possible to use custom SSL certificate for a server

TheSomeMan:


ojousima (approved):
Thanks

PR #761, opened 2023-05-11
Fix #760: tcp_transport: Stack corruption after some network errors

### LINKED ISSUE #760: tcp_transport: Stack corruption after some network errors

TheSomeMan:


ojousima (approved):
Thanks, what was the root cause of stack corruption?

TheSomeMan:
> Thanks, what was the root cause of stack corruption?

The problem only occurred when communicating with the local HTTPS server. 
Normally the client would close the connection first, but in this situation the server's socket close message came first and when handling this error, the function esp_transport_set_errors is called, which should save the error code into the internal structure.
The function esp_transport_set_errors takes an argument of the type pointer to esp_tls_last_error_t, but for SSL connections it actually operates with a bigger structure esp_tls_error_storage_t, declared locally in esp_tls_error_capture.c.
So, esp_transport_set_errors calls esp_tls_get_and_clear_error_type, which casts the original pointer to esp_tls_error_storage_t and modifies the last field of it - 'sock_errno' and at that moment the stack is corrupted.

Issue #476, opened 2022-02-02
Support user SSL certs


ojousima:
Users should be able to upload their self-signed cert to the gateway. The cert should also be updatable with the remote configuration.

hlounent:
X.509 client certificate would be great for authenticating to MQTT server.

I have an MQTT server that forwards data to an InfluxDB instance (and to Grafana from there) but I've currently only enabled client certificate authentication for security reasons and would prefer not to enable password authentication. Client certificates are more secure especially if the CA used for issuing them is stored in an offline system not connected to Internet.

The client should support sending a certificate chain that contains the end-entity client certificate itself and any intermediate CA certificates in order to establish the chain of trust to the CA certificate that the server uses as its trust anchor.

markoaamunkajo:
Recheck once more with Otso

ojousima:
What is needed to be rechecked?

ojousima:
@TheSomeMan is it possible to fetch the certificate with remote configuration? If not, let's assign that feature to backlog and mark this issue as finished.

TheSomeMan:
> is it possible to fetch the certificate with remote configuration? If not, let's assign that feature to backlog and mark this issue as finished.

Yes, client and server certificate checking is supported for the remote configuration.

ojousima:
>> is it possible to fetch the certificate with remote configuration? If not, let's assign that feature to backlog and mark this issue as finished.

> Yes, client and server certificate checking is supported for the remote configuration.

I meant is it possible to fetch the certificate itself from a remote source instead of upload over LAN?

TheSomeMan:
> I meant is it possible to fetch the certificate itself from a remote source instead of upload over LAN?

Yes, it can be implemented, but let's schedule it for v1.15

markoaamunkajo:
@ojousima could you please test SSL's one more time, just to make sure

ojousima:
Changes: 
- Support TLS v1.3 if feasible
- Re-enable check after uploading certificate file
- Support .crt for certificate and .key for private key

markoaamunkajo:
This task is completed

PR #763, opened 2023-05-11
Close #762: Add API to update partitions table - add partition 'gw_cfg_def'

### LINKED ISSUE #762: Update partitions table on boot - add partition 'gw_cfg_def'

TheSomeMan:


ojousima (commented):
How were the partition table bytes generated? How the constant is verified for correctness?

ojousima (commented):
I would like to discuss this a bit further before merge, can we leave this pending until next Wednesday call?

TheSomeMan:
> I would like to discuss this a bit further before merge, can we leave this pending until next Wednesday call?

Sure, this task is needed for #476, but at the last meeting we decided to postpone it.

main/partition_table.c
```diff
@@ -0,0 +1,74 @@
...
+        LOG_WARN("Partition table do not match - need to update");
+        err = spi_flash_erase_range(partition_table_address, SPI_FLASH_SEC_SIZE);
+        if (ESP_OK != err)
+        {
+            LOG_ERR_ESP(err, "%s failed", "esp_flash_erase_region");
+            os_free(p_buf);
+            return false;
+        }
+        err = spi_flash_write(partition_table_address, partition_table_start, partition_table_size);
+        if (ESP_OK != err)
```
ojousima, line 62: If I am understanding correctly, ending in this error is a disaster scenario and Gateway will not boot before flashing because of missing partition table. Could the read-erase-write be written in a loop for trying to recover out of possible error? e.g. 
```
const int table_updated = false;
do{
  table_updated = verify_table();
  if(!table_updated){
    erase_table();
    write_table();
  }
} while(!table_updated)
```
?

ojousima (changes requested):
Overall looks good, but if I'm understanding correctly error in this code will brick the Gateway. Please check the comments and let me know your thoughts about mutex and error handling.

main/partition_table.c
```diff
@@ -0,0 +1,74 @@
...
+#define LOG_LOCAL_LEVEL LOG_LEVEL_INFO
+#include "log.h"
+
+static const char TAG[] = "partition_table";
+
+extern const uint8_t partition_table_start[] asm("_binary_partition_table_bin_start");
+extern const uint8_t partition_table_end[] asm("_binary_partition_table_bin_end");
+
+bool
+partition_table_check_and_update(void)
```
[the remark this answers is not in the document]
TheSomeMan: > Can this function be protected with e.g. mutex to be 100 % sure that there will never be two threads running the function in parallel?

sure, it's a good idea

main/partition_table.c
TheSomeMan: Retrying in case of failure is a good idea, but I think we should limit the number of tries. Maybe 10 attempts would be enough?

TheSomeMan:
> Overall looks good, but if I'm understanding correctly error in this code will brick the Gateway. 

Yes, you're right. And therefore this code will only be called by the user when uploading SSL certificates. I'm going to add a confirmation window to UI with a notification that a potentially unsafe operation needs to be performed.

ojousima (approved):
Thanks

PR #766, opened 2023-05-11
Fix #765: Memory leak in http_download if server returns HTTP error

### LINKED ISSUE #765: Memory leak in http_download if server returns HTTP error

TheSomeMan:


ojousima (commented):
This seems a bit fragile on memory management to me. Where is the memory allocated and why we free it conditionally?

TheSomeMan:
> This seems a bit fragile on memory management to me. Where is the memory allocated and why we free it conditionally?

`http_server_resp_t` structure contains HTTP response code and data to send as HTTP body.
The data can be located in flash memory, in statically allocated memory, on FatFs, or in the heap:
```
typedef enum http_content_location_e
{
    HTTP_CONTENT_LOCATION_NO_CONTENT,
    HTTP_CONTENT_LOCATION_FLASH_MEM,
    HTTP_CONTENT_LOCATION_STATIC_MEM,
    HTTP_CONTENT_LOCATION_HEAP,
    HTTP_CONTENT_LOCATION_FATFS,
} http_content_location_e;
```

When data is located in the heap, we need to free the allocated memory after HTTP response is sent.
But some functions which call http_download_by_handle just check if HTTP status is OK and then return true/false. So, the code which is responsible for freeing memory after sending HTTP response is not called in this case and we need to free the allocated memory in these functions.

The memory for the response is allocated in http_wait_until_async_req_completed in error handling, for example:
```
            if (ESP_ERR_ESP_TLS_CANNOT_RESOLVE_HOSTNAME == err)
            {
                LOG_ERR("%s failed, err=%d (%s)", "esp_http_client_perform", err, "CANNOT_RESOLVE_HOSTNAME");
                const str_buf_t str_buf = str_buf_printf_with_alloc("Failed to resolve hostname");
                return http_server_resp_502_json_in_heap(str_buf.buf);
            }
```
This error message is passed to UI during HTTP target URL validation.

ojousima (approved):
Thanks for explaining the details, ok to merge

PR #784, opened 2023-06-09
Close #783: Decode Ruuvi data formats 5 and 6, show decoded values using API "/history?decode=true"

### LINKED ISSUE #783: Decode Ruuvi data formats 5 and 6, show decoded values using API "/history?decode=true"

TheSomeMan:


ojousima (approved):
Thanks, let's discuss the flag decode default value in Telegram

PR #788, opened 2023-06-13
Fix #777: User gets stuck if http url is empty and tap on check

### LINKED ISSUE #777: User gets stuck if http url is empty and tap on check
Description:

User is trying to configure URL for HTTP but forgot to add address.
If check is pressed, user will get stuck.

https://github.com/ruuvi/ruuvi.gateway_esp.c/assets/50437378/989aee47-c65a-4315-9c79-32eeb2ad115d

TheSomeMan:


ojousima (approved):
Thanks

PR #793, opened 2023-06-20
Fix #792: Gateway stops sending data to the cloud if HTTP connection to the cloud is aborted during configuration mode activation

### LINKED ISSUE #792: Gateway stops sending data to the cloud if HTTP connection to the cloud is aborted during configuration mode activation

TheSomeMan:


ojousima (approved):
Thanks

Issue #755, opened 2023-05-08
Documentation: check that documentation is up to date v 1.14


markoaamunkajo:
Description:

Add new features to documentation
Publish documentation

Quickstart guide
Advanced documentation (Github)

TheSomeMan:
@markoaamunkajo you can check the draft here: https://app.gitbook.com/o/-LsPfwGmkUWZFBz3qkE9/s/-LsPjfnhNj3v37AZtF-d/~/changes/283/ruuvi-gateway-firmware/gw-getting-started

markoaamunkajo:
QA passed

markoaamunkajo:
@TheSomeMan remember to merge when release is rolling out

markoaamunkajo:
Released, moving to done

PR #797, opened 2023-06-29
#755: Update docs: json schemas and examples

TheSomeMan:


schemas/examples/history_with_decoding_df5.json
```diff
@@ -0,0 +1,44 @@
...
+    "gw_mac": "C8:25:2D:8E:9C:2C",
+    "tags": {
+      "F4:1F:0C:28:CB:D6": {
+        "rssi": -51,
+        "timestamp": "1687922293",
+        "data": "0201061BFF990405167546CAC67C000CFFB403E07796A44D5FF41F0C28CBD6",
+        "dataFormat": 5,
+        "temperature": 28.745,
+        "humidity": 45.3050,
+        "pressure": 150812,
```
ojousima, line 14: The pressure seems higher than what DF5 can actually support, I think valid range is 50000 ... 50000 + 2^16-1.

I'd pick a value near 100 000 since that's usually a reasonable value near sea surface level

schemas/ruuvi_history.schema.json
```diff
@@ -1,120 +1,274 @@
...
+                  "examples": [
+                    45,
+                    45.3050
+                  ]
+                },
+                "pressure": {
+                  "type": "integer",
+                  "title": "Pressure in Pa",
+                  "examples": [
+                    150812
```
ojousima, line 107: Same comment on pressure range

ojousima (approved):
Thanks, seems comprehensive. Please see the comments about pressure range before merge

PR #799, opened 2023-06-30
#783: Fix incorrect pressure decoding in API "/history"

TheSomeMan:


ojousima (approved):
Thanks

markoaamunkajo:
Now shows correct Pa values for all pressure readings, considering this is the expected result. QA passed.

PR #819, opened 2023-07-10
Fix #800: nvs_partition_generator returns error: Size exceeds max allowed length 1984 bytes for key gw_cfg_default

### LINKED ISSUE #800: nvs_partition_generator returns error: Size exceeds max allowed length `1984` bytes for key `gw_cfg_default`
Some flags have been removed from gw_cfg_default.json as a workaround to keep within the size allowed.
Removed flags: remote_cfg_use_ssl_client_cert, remote_cfg_use_ssl_server_cert, mqtt_use_ssl_client_cert, mqtt_use_ssl_server_cert.

It looks like a bug in nvs_partition_generator because the allowed length for string values is 4000 bytes: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html#keys-and-values

Need to check if newer versions of ESP-IDF have fixed this problem.
Also, it is possible to store gw_cfg_default.json as a blob.

TheSomeMan:


ojousima (approved):
Thanks

PR #821, opened 2023-07-19
Close #804: Replace cJSON library for json generation with async sequential generator library

### LINKED ISSUE #804: Replace cJSON library for json generation with async sequential generator library
This is needed for HTTP(S) targets and the "/history" API because cJSON tries to allocate a large buffer for the whole json representation and then for the json string. If you have a large number of sensors, this may result in a lack of memory.

TheSomeMan:


ojousima (approved):
Thanks, it's really nice to see memory-efficient code letting us support new features for our users :)

PR #822, opened 2023-07-22
Configure GitHub actions

TheSomeMan:


ojousima (approved):
Thanks, if I'm reading this correctly the build is done on GitHub and artifacts are stored in GitHub as well?

TheSomeMan:
> Thanks, if I'm reading this correctly the build is done on GitHub and artifacts are stored in GitHub as well?

Yes, that's right, now the artifacts are stored on GitHub, but they are also built on the Jenkins server. I removed all actions like executing Google tests and Sonar cloud checks from Jenkins, all that's left is the firmware build. For building/storing artifacts, we can choose between Jenkins or GitHub or keep both.

TheSomeMan:
@markoaamunkajo This task does not involve any code modification. Now formatting checks, running Google Tests, SonarCloud checks are done in GitHub Actions instead of Jenkins. Building firmware binary files is done in both Jenkins and GitHub Actions.

markoaamunkajo:
Nothing to QA

PR #824, opened 2023-07-25
Fix #823: When checking for the presence of firmware binary files, the returned HTTP error code is ignored

### LINKED ISSUE #823: When checking for the presence of firmware binary files, the returned HTTP error code is ignored

TheSomeMan:


ojousima (approved):
Thanks

PR #839, opened 2023-08-16
Fix #838: Sometimes network connection is closed before all data chunks are sent over HTTP

### LINKED ISSUE #838: Sometimes network connection is closed before all data chunks are sent over HTTP

TheSomeMan:


ojousima (approved):
Thanks

PR #846, opened 2023-08-20
Close #775: Data sending interval adjustment for MQTT target

### LINKED ISSUE #775: Data sending interval adjustment for MQTT target
Description:

Data sending interval adjustment setting for MQTT target

TheSomeMan:


main/adv_post.c
```diff
@@ -573,6 +582,7 @@ adv_post_generate_statistics_info(const str_buf_t* const p_reset_info)
     p_stat_info->reset_cnt    = reset_info_get_cnt();
     p_stat_info->p_reset_info = (NULL != p_reset_info) ? &p_reset_info->buf[0] : "";
 
+    // TODO: Do we need to use a separate nonce for stat?
```
ojousima, line 585: It is my understanding that nonces are used to protect against replay-attacks. Since the stat message is separate from data message, using same nonce should not be an issue.

ojousima (approved):
Thanks, if I'm reading this correctly, the periodic MQTT-transmission works by looping through the advertisement table and sending each advertisement as one MQTT-publish?

The consequence of this is that a periodic MQTT transmission puts a hard limit on the number of sensors gateway can relay, as the advertisement table has 100 entries at most? 

This should be clarified to users in UI and documentation if so.

PR #848, opened 2023-08-28
Close #814: Make custom fw update address to affect also automatic updates

### LINKED ISSUE #814: Make custom fw update address to affect also automatic updates
Currently user can point fw update url to custom address on the wizard's updating page but it won't have any effect on from where Gateway is requesting auto updates. Let's allow user to use this custom address also for auto updates. UI to be specified before implementing.

TheSomeMan:


ojousima (approved):
Thanks

PR #850, opened 2023-08-29
Close #803: Implement optional decoding of DF5/DF6 for custom HTTP(S) target

### LINKED ISSUE #803: Implement optional decoding of DF5/DF6 for custom HTTP(S) target

TheSomeMan:


ojousima (commented):
If I am reading this right, this pull request would make it possible to send decoded-only data to Ruuvi's default backend? 

The backend should return an error code in that case, has it been tested?

TheSomeMan:
> If I am reading this right, this pull request would make it possible to send decoded-only data to Ruuvi's default backend?
> 
> The backend should return an error code in that case, has it been tested?

No, this setting affects only custom HTTP.
image

ojousima (approved):
Thanks

PR #852, opened 2023-08-31
Close #774: Data sending interval adjustment for HTTP custom target

### LINKED ISSUE #774: Data sending interval adjustment for HTTP custom target
Description:

Data sending interval adjustment setting for http custom target

TheSomeMan:


ojousima (approved):
Thanks

PR #855, opened 2023-09-02
Close #854: Filter out duplicate messages

### LINKED ISSUE #854: Filter out duplicate messages
Ruuvi sensors can send the same message (with the same "measurementSequenceNumber") multiple times. Such messages are not a big problem for periodic HTTP relaying, but they might be a problem for immediate MQTT relaying because they increase MQTT traffic by transmitting duplicate messages.
This is crucial for #659 to prevent ping-ponging in case of multiple Gateways working in range extender mode.

TheSomeMan:


ojousima (commented):
How does this work on a high level? 

As far as I can see, the code checks if advertisement table already has equivalent data, and if yes, advertisement is discarded. On HTTP transmissions it should not have any effect since only the latest, locked table is sent. Does MQTT somehow rely on the `adv_table` when MQTT is not using a batch mode?

TheSomeMan:
> How does this work on a high level?
> 
> As far as I can see, the code checks if advertisement table already has equivalent data, and if yes, advertisement is discarded. On HTTP transmissions it should not have any effect since only the latest, locked table is sent. Does MQTT somehow rely on the `adv_table` when MQTT is not using a batch mode?

Yes, MQTT now uses `adv_table` in immediate mode as well. It was changed in #775 (commit 06f6f8f23fe47a90e414dd5a768722063177fdbe).
Now when new data is received, EVENT_MGR_EV_RECV_ADV generates signal ADV_POST_SIG_ON_RECV_ADV in `adv_post` task. The handler of this signal sends the advertisement from the head of `adv_table_list3` via MQTT and if this list is not empty the ADV_POST_SIG_ON_RECV_ADV is sent to `adv_post` task immediately.
This solution is more reliable, because in the previous approach, sending advs via MQTT in the parser thread may cause data loss in case of some network delays. 

adv_post.c:
```
adv_post_handle_sig_recv_adv(void)
{
    LOG_DBG("Got ADV_POST_SIG_RECV_ADV");
    if (gw_cfg_get_mqtt_use_mqtt() && (0 == gw_cfg_get_mqtt_sending_interval()) && gw_status_is_mqtt_connected())
    {
        const bool   flag_ntp_use              = gw_cfg_get_ntp_use();
        const time_t timestamp_if_synchronized = time_is_synchronized() ? time(NULL) : 0;
        const time_t timestamp  = flag_ntp_use ? timestamp_if_synchronized : (time_t)metrics_received_advs_get();
        adv_report_t adv_report = { 0 };
        if (adv_table_read_retransmission_list3_head(&adv_report))
        {
            if (mqtt_publish_adv(&adv_report, flag_ntp_use, timestamp))
            {
                adv_post_last_successful_network_comm_timestamp_update();
            }
            else
            {
                LOG_ERR("%s failed", "mqtt_publish_adv");
            }
        }
        if (!adv_table_read_retransmission_list3_is_empty())
        {
            os_signal_send(g_p_adv_post_sig, adv_post_conv_to_sig_num(ADV_POST_SIG_ON_RECV_ADV));
        }
    }
}
```

ojousima (approved):
Thanks. I think the deduplication will fail on installations with over 100 tags in range, but less than 1% of users will encounter that scenario. The special cases can implement filtering on their backend.

TheSomeMan:
> Thanks. I think the deduplication will fail on installations with over 100 tags in range, but less than 1% of users will encounter that scenario. The special cases can implement filtering on their backend.

As long as MQTT transmission speed exceeds the Bluetooth data reception speed, this should work stably on any number of tags.
But if the network is slow, then we will lose data in any case. In the previous solution, we can lose data in the parser thread if MQTT blocks it for a long time. In the current solution, the parser thread will not lose data, but it can overwrite old data.

PR #858, opened 2023-09-10
Refactoring: split big adv_post.c to multiple files

TheSomeMan:


ojousima:
How much effort it would be to add test coverage to old file / new files? If the SonarCloud reported 0.2 % is right, it's difficult to catch is something has broken in this commit. 

My approach here would be to write tests for the old file, at least on the most used code paths and then add the refactor. If writing the tests is a big effort and you are confident in the refactor we could merge this as-is. Please let me know your thoughts on the approach.

ojousima (approved):
Thanks, in all honesty this is too big pull request for me to review carefully but as it's adding library files and tests for a refactor it should be good to go

PR #859, opened 2023-09-10
Fix #857: Increase stack size for MQTT

### LINKED ISSUE #857: Increase stack size for MQTT

TheSomeMan:


ojousima (approved):
Thanks, possibly (6U * 1024U) to keep the style of above #define and old #define value as unsigned.

PR #871, opened 2023-09-17
Fix #868: Problem with Gateway configuration via Nginx proxy manager

### LINKED ISSUE #868: Problem with Gateway configuration via Nginx proxy manager
Kirbo  [5:32 PM]
Hi, I just found out I cannot login into the Ruuvi Gateway behind a proxy. I think this might have been done on purpose, but "as a user, I'd like to be enabled to allow login from behind proxy", i.e. allow login from behind proxy/allow login from list of origins/idk.
I'm currently using Nginx Proxy Manager (https://nginxproxymanager.com/), which both adds SSL certificates to my local URLs and redirects e.g. https://ruuvi.hima.osoite.com/ into http://192.168.1.130/ (so I don't need to remember the IP-addresses and ports for like 20 different local services that I'm currently running (which pretty much every one of them uses a different port: 80, 81, 443, 3000, 4408, 4409, 8006, 8080, 8081, 8123, 8888, 9000, 10086...).
The login itself opens up as it should, but the response I get from the POST /auth endpoint is 401. If I open the login via the direct IP-address for the Gateway, I am able to login without issues.
Disclaimer: I could also have wrong settings in "NPM", but here are few screenshots anyway, if it helps you somehow
(..and no, the URLs are not accessible outside of my LAN, but still wanted to redact them. In case anyone is interested, my router advertises for LAN clients to use the AdGuard Home as DNS-server, which has a wildcard custom DNS-rewrite for *.[hima.osoite.com](http://hima.osoite.com/) -> 192.168.1.9, which is the IP-address for the "NPM", which then uses the Proxy Pass to redirect the [ruuvi.hima.osoite.com](http://ruuvi.hima.osoite.com/) -> 192.168.1.130:80)

TheSomeMan:


ojousima (approved):
Thanks

PR #873, opened 2023-09-19
Fix #872: Too many log messages "Can't send advs, the time is not yet synchronized"

### LINKED ISSUE #872: Too many log messages "Can't send advs, the time is not yet synchronized"

TheSomeMan:


ojousima (approved):
Thanks

PR #875, opened 2023-09-19
Fix #874: Incorrect LED notification for custom HTTP target about network connection problem

### LINKED ISSUE #874: Incorrect LED notification for custom HTTP target about network connection problem

TheSomeMan:


ojousima (approved):
Thanks

PR #882, opened 2023-10-06
Fix #881: Task watchdog triggers for various threads while adv_post_task establishing SSL connection with some HTTPS servers

### LINKED ISSUE #881: Task watchdog triggers for various threads while adv_post_task establishing SSL connection with some HTTPS servers

TheSomeMan:


ojousima (approved):
Thanks

PR #891, opened 2023-10-17
Close #887: High-load continuous data retransmission causes excessive logs during reconfiguration, triggering task watchdog

### LINKED ISSUE #887: High-load continuous data retransmission causes excessive logs during reconfiguration, triggering task watchdog.

TheSomeMan:


ojousima (approved):
Thanks

PR #892, opened 2023-10-17
Fix #888: nRF52 post-reboot immediate data output triggers 'assert' causing reboot loop

### LINKED ISSUE #888: nRF52 post-reboot immediate data output triggers 'assert' causing reboot loop

TheSomeMan:


ojousima (approved):
Thanks, is this critical enough to warrant a v1.14.4 release?

Issue #889, opened 2023-10-17
Optimize message parsing received via UART from nRF52


TheSomeMan:
[Refactor messaging parsing: Use single-threaded state machine in place of two-threaded implementation](https://github.com/ruuvi/ruuvi.comm_tester.c/issues/28)

This issue relates to #886.

The change is made for multiple reasons:

1. Performance Improvements.
2. Simpler debugging due to the reduction of synchronization issues.
3. Increased maintainability.
4. Memory efficiency: This implementation contributes towards saving approximately 6.5 kB of memory.

PR #903, opened 2023-10-25
[#889] Update ruuvi.comm_tester.c to the master branch, increase stack size for rx_uart_task from 2 to 4kB

TheSomeMan:


ojousima (approved):
Thanks

PR #904, opened 2023-10-26
Fix #901: mqtt_client incorrectly handles OUTBOX_MAX_SIZE and allows unlimited buffer growth when using QoS1 or QoS2

### LINKED ISSUE #901: mqtt_client incorrectly handles OUTBOX_MAX_SIZE and allows unlimited buffer growth when using QoS1 or QoS2

TheSomeMan:


ojousima (approved):
Thanks. If this can be contributed upstream to Espressif, please open a PR with this patch on their end as well.

PR #906, opened 2023-10-26
Close #905: MQTT: set default QoS to 0 to disable message buffering

### LINKED ISSUE #905: MQTT: set default QoS to 0 to disable message buffering

TheSomeMan:


ojousima:
How does this work when MQTT is throttled? I think a message with buffered sensor reads could be sent with QOS 1

TheSomeMan:
> How does this work when MQTT is throttled? I think a message with buffered sensor reads could be sent with QOS 1

When QoS0 mode is used, packets are simply sent over MQTT and are not placed in the internal buffer. In this case, data delivery is guaranteed by the TCP layer (MQTT runs on top of TCP). 
When MQTT is throttled, I suppose that one of two things could happen:
- some TCP packets will be automatically resent by TCP layer 
- TCP connection will be closed and MQTT will reconnect
When using QoS1 mode, MQTT adds additional protection over TCP layer and checks that MQTT ACK has arrived from the server for each MQTT packet transmitted and if no ACK arrives, the data is retransmitted from the internal buffer.
QoS1 mode may be required if the server cannot process the incoming data stream in time and discards some of it. In this case, the task of buffering is shifted from the server side to MQTT-client. In our case, Ruuvi Gateway has very limited memory resources and can buffer only about 10 MQTT packets when QoS1 is used, which is practically useless.

ojousima (approved):
Thanks

PR #911, opened 2023-11-01
Close #909: Move MQTT data sending from adv_post thread to a separate thread

### LINKED ISSUE #909: Move MQTT data sending from adv_post thread to a separate thread

TheSomeMan:


ojousima (approved):
Thanks

PR #916, opened 2023-11-16
Close #789: Use Ruuvi Time Server as a last option time.ruuvi.com

### LINKED ISSUE #789: Use Ruuvi Time Server as a last option time.ruuvi.com
Description:

User wouldn't need to make adjustments to firewall other than ruuvi.com
We would keep current server options as secondary etc. as backup.

Would simplify initial configuration for user when no need to grant access to elsewhere beyond ruuvi.com

To be discussed in meeting

TheSomeMan:


ojousima (approved):
Thanks, I set up 
```
ntp.ruuvi.com
time.ruuvi.com
```
Which point to the same server. According to https://network-tools.webwiz.net/ntp-server-test.htm the ntp server works.

markoaamunkajo:
QA passed

PR #918, opened 2023-11-16
Close #917: Add a 1 second delay before sending a response to an authentication request to prevent brute force attacks

### LINKED ISSUE #917: Add a 1 second delay before sending a response to an authentication request to prevent brute force attacks

TheSomeMan:


ojousima (approved):
Thanks

PR #919, opened 2023-11-17
Fix #897: Web Configurator used via Ethernet loses connection to the Gateway after 1 hour

### LINKED ISSUE #897: Web Configurator used via ethernet loses connection to the Gateway after 1 hour
Description:

I was in a call to discuss about UI elements and stayed long time in configurator, over an 1 hour. Suddenly the configurator showed wifi connection was lost, but I was not connected through wifi and I'm assuming internet didn't hang.

TheSomeMan:


ojousima (approved):
Thanks

markoaamunkajo:
QA passed

PR #926, opened 2023-11-22
#925: Gateway reboots during URL validation if authentication is used but token or username is empty

TheSomeMan:


ojousima (approved):
Thanks

markoaamunkajo:
QA passed

Issue #927, opened 2023-11-22
Error description with the message from the server during URL validation contains excessive escaping


TheSomeMan:

TheSomeMan:
In order to test it, you need to enable custom HTTPS and try to check URL "https://network.ruuvi.com/record2". The server will return an error message like this:
```
{"message":"Not Found"}
```

This message should be properly escaped when displaying in UI:
```
HTTP response status: 404, Message from the server: {"message":"Not Found"}
```


The previous build generated extra escaping like this:
```
HTTP response status: 404, Message from the server: {\"message\":\"Not Found\"}
```

markoaamunkajo:
QA passed

PR #928, opened 2023-11-22
#927: Error description with the message from the server during URL validation contains excessive escaping

TheSomeMan:


ojousima (approved):
Thanks

PR #931, opened 2023-11-24
Fix #930: Configure button press event is sometimes lost, resulting in the inability to reset settings

### LINKED ISSUE #930: Configure button press event is sometimes lost, resulting in the inability to reset settings

TheSomeMan:


ojousima (approved):
Thanks

PR #932, opened 2023-11-30
Close #922: Enable WPS to automatically configure Wi-Fi

### LINKED ISSUE #922: Enable WPS to automatically configure Wi-Fi

TheSomeMan:


ojousima (approved):
Thanks, please check with @markoaamunkajo and @laurijamsa that the led blinking sequences with WPS get documented and included in our quick start guide.

PR #934, opened 2023-12-04
Fix #933: Gateway sends two messages to server without delay, which causes "Gateway throttled" response

### LINKED ISSUE #933: Gateway sends two messages to server without delay, which causes "Gateway throttled" response

TheSomeMan:


ojousima (approved):
Thanks

PR #936, opened 2023-12-04
Fix #935: Static IP configuration for Ethernet is not saved

### LINKED ISSUE #935: Static IP configuration for Ethernet is not saved

TheSomeMan:


ojousima (approved):
Thanks

PR #947, opened 2023-12-14
Fix #946: Sometimes reconnecting to Wi-Fi does not work correctly

### LINKED ISSUE #946: Sometimes reconnecting to Wi-Fi does not work correctly

TheSomeMan:


ojousima (approved):
Thanks

ojousima (approved):
Thanks

PR #955, opened 2023-12-20
Fix #954: Green LED can erroneously turn off when there is no communication with one of multiple configured destinations

### LINKED ISSUE #954: Green LED can erroneously turn off when there is no communication with one of multiple configured destinations

TheSomeMan:


ojousima (approved):
Thanks

PR #957, opened 2023-12-21
Fix #956: Wi-Fi AP stops before sending HTTP response when remote configuration download is enabled

### LINKED ISSUE #956: Wi-Fi AP stops before sending HTTP response when remote configuration download is enabled

TheSomeMan:


ojousima (approved):
Thanks

PR #961, opened 2023-12-29
Fix #960: UI alignment fixes when FI translation is enabled

### LINKED ISSUE #960: UI alignment fixes when FI translation is enabled

TheSomeMan:


ojousima (approved):
Thanks

PR #968, opened 2024-01-18
Close #967: Implement TLS v1.3 support (port from ESP-IDF v5.1.2)

### LINKED ISSUE #967: Implement TLS v1.3 support (port from ESP-IDF v5.1.2)

TheSomeMan:


ojousima (commented):
There's no realistic way of reviewing 2171 files updated.

Is it possible to add the component as a submodule pointing to upstream somehow?

TheSomeMan:
> There's no realistic way of reviewing 2171 files updated.
> 
> Is it possible to add the component as a submodule pointing to upstream somehow?

"mbedtls" is a submodule, so it can be added as a submodule. 
But wpa_supplicant, tcp_transport and protocomm are not submodules. I can split the changes into two steps - first create a local copy of the components, then make the changes in a separate PR. Or I can try to port the whole project to ESP-IDF v5.1.2 again, now it should be easier since most of the patches were ported.

ojousima (approved):
Let's merge this as-is when ready, it doesn't make a functional difference if we add the files through a more complex process.

PR #971, opened 2024-01-30
Fix #969: Gateway sometimes reboots due to erroneous detection of memory leaks

### LINKED ISSUE #969: Gateway sometimes reboots due to erroneous detection of memory leaks (v1.14.3)
[err_detection_mem_leak.txt](https://github.com/ruuvi/ruuvi.gateway_esp.c/files/14099359/err_detection_mem_leak.txt)

```
I (16750) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 66548
I (19493) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 67952
I (22285) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 67956
I (25184) ruuvi_gateway: [main/1] free heap: 13116
I (26428) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 67980
I (32218) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 67532
I (36791) ruuvi_gateway: [main/1] free heap: 19476
I (42194) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 69404
I (48184) ruuvi_gateway: [main/1] free heap: 19080
I (52214) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 68100
I (59779) ruuvi_gateway: [main/1] free heap: 20472
I (62191) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 66388
I (72206) ADV_POST_TASK: [adv_post_task/5] Cur free heap: 64796
I (72284) ruuvi_gateway: [main/1] free heap: 20128
E (72284) ruuvi_gateway: [main/1] ./main/main_loop.c:208 {main_task_handle_sig_log_heap_usage}: Only 19KiB of free memory left - probably due to a memory leak. Reboot the Gateway.
```

TheSomeMan:


ojousima:
Do we  detect allocation failures somewhere? Or does the max_free_heap limit guarantee that there is enough memory for normal operation?

TheSomeMan:
> Do we detect allocation failures somewhere? Or does the max_free_heap limit guarantee that there is enough memory for normal operation?

In many cases, allocation failure is not critical. For example, if there is not enough memory while sending data to the HTTP target, we can just retry this operation.
Yes, max_free_heap limit guarantees that it's enough to perform data sending over HTTPS. And if we have less memory than 50KB, then it is highly likely that a memory allocation error will occur when sending data. So, if the amount of memory in the system for a long time is less than this limit, there may be a memory leak somewhere and it is safer to reboot the gateway so that HTTPS/MQTTS connections are guaranteed to open.

ojousima (approved):
Thanks

PR #980, opened 2024-02-07
Fix #978, #967, #760: tcp_transport: Stack smashing protect failure after some network errors, improve logging in esp_tls and tcp_transport

### LINKED ISSUE #978: tcp_transport: Stack smashing protect failure after some network errors
Original issue #760, fixed in #761.
The problem appeared again after implementing TLS v1.3 support (#967) because not all places in the tcp_transport component were patched.

TheSomeMan:


components/esp-tls/esp_tls.c
```diff
@@ -308,17 +313,20 @@ static esp_err_t esp_tls_set_socket_non_blocking(int fd, bool non_blocking)
...
     }
     return ESP_OK;
 }
 
-static inline esp_err_t esp_tls_tcp_connect(const ip_addr_t* const p_remote_ip, int port, const esp_tls_cfg_t *cfg,
+static esp_err_t esp_tls_tcp_connect(const ip_addr_t* const p_remote_ip, int port, const esp_tls_cfg_t *cfg,
             esp_tls_error_handle_t error_handle, int *sockfd)
 {
+    char remote_ip_str[IP4ADDR_STRLEN_MAX];
+    ip4addr_ntoa_r(&p_remote_ip->u_addr.ip4, remote_ip_str, IP4ADDR_STRLEN_MAX);
```
ojousima, line 328: Is this safe on IPv6 only servers? 

It seems to me that `ip4addr_ntoa_r` calls `ip4_addr_get_u32` which dereferences given argument. Since the argument is the address of `p_remote_ip->u_addr.ip4` this *seems* like even in case there is no ipv4 address available, it would just point to some uninitialized memory in a worst case and it would be used as a readonly.

ojousima (approved):
Thanks. 

Was the fix in components/tcp_transport/transport_ssl.c around line 290 with error handling? Is that a fix we could contribute upstream?

TheSomeMan:
> Was the fix in components/tcp_transport/transport_ssl.c around line 290 with error handling? 
> Is that a fix we could contribute upstream?

Yes, the fix is here. It was done in a separate commit [bd1b986f4f32af51c34c61bd81ae04fa3d27e6f7].
But there is nothing to contribute upstream. I unintentionally made this bug when porting my previous patches to tcp_transport from ESP-IDF v5.1.2.

components/esp-tls/esp_tls.c
TheSomeMan: > Is this safe on IPv6 only servers?
> 
> It seems to me that `ip4addr_ntoa_r` calls `ip4_addr_get_u32` which dereferences given argument. Since the argument is the address of `p_remote_ip->u_addr.ip4` this _seems_ like even in case there is no ipv4 address available, it would just point to some uninitialized memory in a worst case and it would be used as a readonly.

No, some patches are not IPv6 compatible. But IPv6 support is disabled in this project.
Do you think IPv6 support should be enabled?  Could this be planned for a future version 1.16.x, what do you think?

ojousima: Let's leave the IPv6 support for 1.16 or some other future support in this case, so far we have not had a user request it.

PR #987, opened 2024-02-08
Fix #986: mqtt_client: incorrect error code is printed to log if esp_transport_read returns error

### LINKED ISSUE #986: mqtt_client: incorrect error code is printed to log if esp_transport_read returns error

TheSomeMan:


ojousima (approved):
Thanks

PR #993, opened 2024-02-09
Fix #992: UI: MQTT: Changing the state of "Disable the use of retained messages" checkbox should enable the "Check" button

### LINKED ISSUE #992: UI: MQTT: Changing the state of "Disable the use of retained messages" checkbox should enable the "Check" button

TheSomeMan:


ojousima (approved):
Thanks

PR #997, opened 2024-02-13
Close #996: Implement support for configuring the size of TLS connection I/O buffers to prevent problems with downloading firmware updates

### LINKED ISSUE #996: Implement support for configuring the size of TLS connection I/O buffers to prevent problems with downloading firmware updates

TheSomeMan:


TheSomeMan:
To test it you need to check logs and find the following lines:
```
Configure size of TLS I/O buffers: in_content_len=8192, out_content_len=4096
```
or
```
Configure size of TLS I/O buffers: in_content_len=16384, out_content_len=4096
```

PR #999, opened 2024-02-14
Close #998: Release saved TLS session ticket as soon as it is used

### LINKED ISSUE #998: Release saved TLS session ticket as soon as it is used

TheSomeMan:


ojousima (approved):
Thanks

PR #1004, opened 2024-02-16
Fix #1003: Red LED starts blinking for a few seconds after checking if the new configuration is available on the remote configuration server

### LINKED ISSUE #1003: Red LED starts blinking for a few seconds after checking if the new configuration is available on the remote configuration server

TheSomeMan:


ojousima (approved):
Thanks

PR #1005, opened 2024-02-16
[#998] Fix a problem with premature displacement of a stored TLS session ticket from the buffer

TheSomeMan:


ojousima (approved):
Thanks

TheSomeMan:
To test it, you need to configure your HTTPS server, check that the gateway successfully establishes the HTTPS connection, try to restart the server and check that HTTP POST works as expected.

markoaamunkajo:
QA passed

PR #1007, opened 2024-02-16
Fix #1006: esp_http_client: Opening HTTP connection leads to leakage of open sockets

### LINKED ISSUE #1006: esp_http_client: Opening HTTP connection leads to leakage of open sockets

TheSomeMan:


ojousima (approved):
Thanks

PR #1013, opened 2024-02-26
[#967] Fix issues related to porting TLS v1.3 from ESP-IDF v5.1.2 

TheSomeMan:


ojousima:
How much of the code is your original work and how much is ported from some external component? Especially in `components/tcp_transport/transport_ssl.c`? Is some of the work reasonable to contribute upstream, e.g. null ptr check`if (cfg && cfg->non_block)`?

TheSomeMan:
> How much of the code is your original work and how much is ported from some external component? Especially in `components/tcp_transport/transport_ssl.c`? Is some of the work reasonable to contribute upstream, e.g. null ptr check`if (cfg && cfg->non_block)`?

The changes in these components are quite large and it will be quite difficult to contribute them to upstream. As for the `if (cfg && cfg->non_block)`, this part was added by me into esp_tls.c to track timeout expiration in asynchronous mode when establishing a connection. In the original implementation, even in asynchronous mode, execution was blocked for 10 seconds on the `select` call when establishing a connection, which resulted in task watchdog triggering.

In transport_ssl.c the main changes are as follows:
* Added support for saving TLS session tickets and reusing them for fast re-establishing HTTPS connection
* Fixed bug in ssl_read/ssl_write - these functions even in asynchronous mode blocked execution on select
* Added a timer to track timeout expiration in asynchronous mode for ssl_read/ssl_write
* The same bug was fixed and the timer was added for tcp_read/tcp_write
* Added a function to dynamically configure SSL content buffer size

In esp_tls.c the changes are as follows:
* Implemented asynchronous hostname resolution
* Fixed the problem with establishing a connection in asynchronous mode - it blocks for 10 seconds on `select`
* Added timeout handling during the TLS handshake (if the server doesn't respond, it could wait forever)
* Fixed a bug in `esp_tls_plain_tcp_connect` and `esp_tls_conn_new_sync`, they used `gettimeofday` for timeout tracking, which causes problems during initial time synchronization
* Improved logging by adding task name, function name, socket number, and hostname to each log message

ojousima (approved):
Thanks

TheSomeMan:
To test it you need to configure various types of relaying: HTTP/HTTPS, MQTT/MQTTS/WS/WSS and check that it works as expected.

ojousima:
WS/WSS not tested

PR #1029, opened 2024-04-16
Fix #1019: The network watchdog timer timeout is disabled when configuration mode is activated

### LINKED ISSUE #1019: The network watchdog timer timeout is disabled when configuration mode is activated
When the browser opens the Configuration Wizard, the gateway enters configuration mode. While the browser is active, the gateway does not relay data. Gateway must be rebooted by the network watchdog timeout after 1 hour (in case the user forgot to close the browser with the configuration wizard).

TheSomeMan:


ojousima (approved):
Thanks

Issue #1022, opened 2024-04-08
Gateway reboots with IntegerDivideByZero when trying to update firmware from server that does not return 'Content-length'


TheSomeMan:
```
I (566687) transport_base: [fw_update_task] [<NULL>] Configure size of TLS I/O buffers: in_content_len=16384, out_content_len=4096
I (566700) esp-tls: [fw_update_task] esp_tls_init: tls=0x3ffe0b04
I (566758) esp-tls: [fw_update_task] Hostname 'my-https-server.local' resolved to 192.168.1.145
I (566809) esp-tls: [fw_update_task] esp_tls_tcp_connect: Open socket 54 for 192.168.1.145:8000
I (566814) http_download: [fw_update_task/1] HTTP_EVENT_ON_CONNECTED
I (566817) http_download: [fw_update_task/1] HTTP_EVENT_HEADER_SENT
I (566824) http_download: [fw_update_task/1] HTTP_EVENT_ON_HEADER, key=Content-type, value=application/octet-stream
I (566833) ADV_POST_TASK: [adv_post_task/5] Advs cnt: 7
I (566839) http_download: [fw_update_task/1] HTTP_EVENT_ON_HEADER, key=Cache-Control, value=no-store, no-cache, must-revalidate, max-age=0
I (566852) http_download: [fw_update_task/1] HTTP_EVENT_ON_HEADER, key=Pragma, value=no-cache
I (566861) fw_update: [fw_update_task/1] Write to OTA-partition ota_1, offset 0, size 1908
Guru Meditation Error: Core  0 panic'ed (IntegerDivideByZero). Exception was unhandled.

Core  0 register dump:
PC      : 0x400e093a  PS      : 0x00060630  A0      : 0x800ed36a  A1      : 0x3ffe9dd0  
A2      : 0x3ffe6f80  A3      : 0x00000774  A4      : 0x0000ba54  A5      : 0x00000000  
A6      : 0x00000000  A7      : 0x3ffea1ec  A8      : 0x00009510  A9      : 0x3ffe9d80  
A10     : 0x00000003  A11     : 0x000174a8  A12     : 0x3f408dcc  A13     : 0x8ed73ab6  
A14     : 0x3ffe9d80  A15     : 0x3ffe60a0  SAR     : 0x00000004  EXCCAUSE: 0x00000006  
[...]

PR #1039, opened 2024-05-05
[#1022] Fix bug in esp_http_client_perform - HTTP HEAD request may fail due to race condition on slow server if server does not send HTTP Content-Length header

TheSomeMan:


ojousima (approved):
Thanks

ojousima (approved):
Thanks

PR #1042, opened 2024-05-06
Fix #1025: Gateway erroneously forced to use TLS v1.3, preventing it from establishing a connection using TLS v1.2, resulting in SSL_HANDSHAKE_FAILED when connecting to Azure IoT Hub

### LINKED ISSUE #1025: Gateway erroneously forced to use TLS v1.3, preventing it from establishing a connection using TLS v1.2, resulting in SSL_HANDSHAKE_FAILED when connecting to Azure IoT Hub
Does anyone else experience ESP_ERR_MBEDTLS_SSL_HANDSHAKE_FAILED when connecting to Azure IoT Hub?
This does not happen on v1.14.3

TheSomeMan:


ojousima (approved):
Thanks

PR #1046, opened 2024-05-08
Close #1044: Add HTTP authentication method "API key" that does not prefix the key with the word "Token"

### LINKED ISSUE #1044: Add HTTP authentication method "API key" that does not prefix the key with the word "Token"
Hi,

Problem as in the title - the "API key (token)" is prefixed unexpectedly by word "Token" which makes impossible to authenticate against IoT Hub/IoT Central with SAS (shared access signature) key which is set in Authorization header.

Is the "Token" required by any specific protocol? If so, one could simply type the word "Token" if needed in front of the actual token.

Please note that this is BC breaking change, so it needs to be properly announced. Alternatively [a new, fully customizable authentication method](https://github.com/ruuvi/ruuvi.gateway_esp.c/issues/1045) could be implemented.

TheSomeMan:


ojousima (approved):
Thanks, please also update the docs with an example

TheSomeMan:
> Thanks, please also update the docs with an example

Yes, I've updated the documentation.

TheSomeMan:
@markoaamunkajo to test this, you can run the script `./ruuvi.gwui.html/scripts/http_server_auth.py`, configure the gateway to send data to your computer 'http://<YOUR_IP>:8000/record' and check HTTP header "Authorization" in the script console.

markoaamunkajo:
QA passed

PR #1054, opened 2024-05-13
Fix #1053: ESP_ERR_MBEDTLS_SSL_HANDSHAKE_FAILED on HTTP endpoint validation "*.logic.azure.com" because new random number is generated on second "Client Hello"

### LINKED ISSUE #1053: ESP_ERR_MBEDTLS_SSL_HANDSHAKE_FAILED on HTTP endpoint validation "*.logic.azure.com" because new random number is generated on second "Client Hello"
Hello,

I am getting `Network error when communicating with the server, err=32794, description=ESP_ERR_MBEDTLS_SSL_HANDSHAKE_FAILED` when validating azure logic apps as an HTTP endpoint

image

Affected firmware version: **v1.15.x**
Works fine on **v1.14.3**

Steps to reproduce:
1. Create the simplest logic app with only two blocks as below:
image
2. rename the first block to something very short (like "ruuvi" or shorter) otherwise you'll run into [problem described here](https://github.com/ruuvi/ruuvi.gateway_esp.c/issues/1052) (because the name becomes a part of the URL making it longer than 255 characters)
3. save the app
4. once the app is saved, copy an HTTP URL of the app from the first block
5. enter the URL as the HTTP endpoint for ruuvi gateway
6. click "Check" button to see the `ESP_ERR_MBEDTLS_SSL_HANDSHAKE_FAILED ` error

It is **not a blocker for me ATM** - just letting you know as it indicates that potentially there is still a problem after introducing TLS 1.3 support. Please note that according to [this report](https://www.ssllabs.com/ssltest/analyze.html?d=prod-16.norwayeast.logic.azure.com&hideResults=on), the logic apps endpoint I am testing against, supports TLS 1.3.

TheSomeMan:
Port commit "Do not generate new random number while receiving HRR": https://github.com/Mbed-TLS/mbedtls/commit/35178fe7ecd473328248c3f0cc0dc3a0603d2a21

### DIFF
--- a/components/mbedtls/mbedtls/library/ssl_client.c
+++ b/components/mbedtls/mbedtls/library/ssl_client.c
@@ -820,10 +820,15 @@ static int ssl_prepare_client_hello(mbedtls_ssl_context *ssl)
         (ssl->handshake->cookie == NULL))
 #endif
     {
-        ret = ssl_generate_random(ssl);
-        if (ret != 0) {
-            MBEDTLS_SSL_DEBUG_RET(1, "Random bytes generation failed", ret);
-            return ret;
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+        if (ssl->handshake->hello_retry_request_count == 0)
+#endif
+        {
+            ret = ssl_generate_random(ssl);
+            if (ret != 0) {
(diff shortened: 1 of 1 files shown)


ojousima (approved):
Thanks

PR #1068, opened 2024-09-27
Close #1069: Add support for BLE extended advertisements

### LINKED ISSUE #1069: Add support for BLE extended advertisements

TheSomeMan:


main/nrf52fw.c
```diff
@@ -555,11 +554,36 @@ nrf52fw_flash_write_firmware(
             return false;
         }
     }
-    if (!nrf52fw_write_current_fw_ver(p_fw_info->fw_ver.version))
+    /*
+     * Starting from gateway_nrf v2.0.0 the firmware image.hex file contains the firmware version in the last segment,
```
ojousima, line 558: Does something ensure that firmware version will get written only in case of successful write of a new firmware? e.g. UICR registers are on lower end of firmware .hex, get written in, power outage occurs in middle of nRF update and UICR is already written. How does GW detect corrupted nRF binary in such case?

ojousima: Answering myself, it seems to me that .hex is not blindly written as-is, but the version code in nRF52 partition is checked, if version matches what GW expects the GW writes hex segments from the file and on success the nRF version hardcoded in ESP code is written. Did I understand correctly?

ojousima (approved):
Thanks, please confirm nRF UICR version update comment/question before merge

main/nrf52fw.c
TheSomeMan: > Does something ensure that firmware version will get written only in case of successful write of a new firmware?

Yes, the memory segments are written sequentially and verified after writing. So if a segment is written with an error, the function will terminate before attempting to write the version to UICR. 
It is expected that the segment for UICR must be placed at the end of the hex file. But this segment is optional - the version is added only for a released firmware with a git tag. If .hex does not contain a segment with the version we need to flash it forcibly (this is also to ensure compatibility with v1.0.0).
So, it tries to write UICR if it's empty (it's not present in .hex). And if it's not empty, then compare it with the expected version and return an error if they don't match.

Here is the code:
image

PR #1071, opened 2024-11-04
Close #1070: Request to HTTP API /history returns error 401 just after reboot if LAN access without a password is configured

### LINKED ISSUE #1070: Request to HTTP API /history returns error 401 just after reboot if LAN access without a password is configured
There is a workaround - open /auth page once after reboot.

TheSomeMan:


ojousima (approved):
Thanks

PR #1074, opened 2025-05-21
Close #1073: Add ACK handling from nRF52, force reboot of nRF52 when no ACK is present

### LINKED ISSUE #1073: Add ACK handling from nRF52, force reboot of nRF52 when no ACK is present

TheSomeMan:


ojousima (approved):
Thanks

PR #1076, opened 2025-05-21
Close #1075: UI: Fix: Dynamic setting of BLE scan filter does not work

### LINKED ISSUE #1075: Dynamic setting of BLE scan filter does not work

TheSomeMan:


ojousima (approved):
Thanks

PR #1099, opened 2025-12-18
Close #1098: Generate signed images of FatFS partitions with nRF52 firmware and GWUI

### LINKED ISSUE #1098: Generate signed images of FatFS partitions with nRF52 firmware and GWUI

TheSomeMan:


ojousima (approved):
Thanks

PR #1108, opened 2025-12-26
Close #1104: Use non-modifiable and reproducible FAT-FS images for GWUI and nRF52 firmware so that signatures can be verified

### LINKED ISSUE #1104: Use non-modifiable FAT-FS images for GWUI and nRF52 firmware so that signatures can be verified

TheSomeMan:


ojousima (approved):
Thanks. I think "Use reproducible FAT-FS..." would be more accurate title than "Use non-modifiable FAT-FS..."?

TheSomeMan:
> Thanks. I think "Use reproducible FAT-FS..." would be more accurate title than "Use non-modifiable FAT-FS..."?

Regular ESP32 FAT-FS images created with mkfatfs use wear-leveling and ESP-IDF v4.2.2 does not support mounting them in read-only mode. So, such images are modifiable on wear-leveling layer, although they are not modified on FAT-FS layer.
Switching to raw FAT-FS images created with mtools allow to create non-modifiable images. 
And special instruction `export SOURCE_DATE_EPOCH=315532800` allows to create reproducible builds because it replaces all file timestamps with this base time.

So, I renamed the PR to "Use non-modifiable and reproducible ..."

PR #1110, opened 2025-12-26
Fix #1109: Fix problem with showing progress of partition erasing after firmware update failure

### LINKED ISSUE #1109: Fix problem with showing progress of partition erasing after firmware update failure

TheSomeMan:


ojousima (approved):
Thanks

PR #1120, opened 2025-12-29
Close #1119: Add SHA256 hash verification of nRF52 firmware on the target MCU

### LINKED ISSUE #1119: Add SHA256 hash verification of nRF52 firmware on the target MCU

TheSomeMan:


ojousima (approved):
Thanks

PR #1122, opened 2025-12-31
Close #1121: Automatically recover FAT-FS partition wiht nRF52 firmware if it was damaged by wear leveling after update

### LINKED ISSUE #1121: Automatically recover FAT-FS partition wiht nRF52 firmware if it was damaged by wear leveling after update

TheSomeMan:


ojousima (approved):
Thanks

PR #1124, opened 2026-01-10
Close #1092: Restart gateway in case of error "emac_esp32: emac_esp32_transmit(223): insufficient TX buffer size"

### LINKED ISSUE #1092: emac_esp32: emac_esp32_transmit(223): insufficient TX buffer size (again)
The symptoms are more or less same as with https://github.com/ruuvi/ruuvi.gateway_esp.c/issues/915 but considering that was fixed ages ago, this is probably something else. The advertisement rate/number of tags has also almost doubled since last time. All of a sudden I started getting those same errors spammed in logs:

Image

and as last time, the gateway appeared offline (no MQTT connection, `/metrics` unavailable, etc). Unlike last time I was sleeping at the time so I was not there to reboot it, but it rebooted on its own after about an hour. Don't know if it's relevant at all, but metrics revealed varying fragmentation with the free memory, which was not present after the reboot:

Image

TheSomeMan:


ojousima (approved):
Thanks

PR #1127, opened 2026-01-20
Close #1126: Add support for data format 7 (motion sensors)

### LINKED ISSUE #1126: Add support for data format 7 (motion sensors)

TheSomeMan:


ojousima (approved):
Thanks, please note that DF 7 will probably evolve a bit before it's final

PR #1148, opened 2026-02-10
Fix #1146: Writing beyond SSL output buffer when clearing memory after closing a TLS connection

### LINKED ISSUE #1146: Writing beyond SSL output buffer when clearing memory after closing a TLS connection

TheSomeMan:
The bug occurs when mbedtls_ssl_setup allocates less memory for the buffer than is necessary because it does not account for header and payload overhead. Consequently, when clearing memory during deinitialization upon closing a TLS connection, writing outside the buffer may occur, thereby corrupting foreign memory.

Introduces helper functions to encapsulate the logic for retrieving and calculating SSL buffer lengths, improving code readability and maintainability.

This change replaces direct access to content length variables with calls to the new helper functions, ensuring consistent buffer size calculations throughout the SSL module.
The helper functions consider both fixed and variable buffer length configurations.


ojousima (approved):
Thanks. 

Could this patch be contributed upstream?

PR #1253, opened 2026-04-03
Close #1093: Increase self-update delay after boot to 2 hours

### LINKED ISSUE #1093: Increase self-update delay after boot to 2 hours
Current Gateway inventory ships with v1.15.3 and at 40 minute mark it updates to v1.16.2

This has caused some confusion with users, as they do other setup and see that Gateway starts blinking red while updating nRF52.

Users do resets, try to connect to configuration hotspot, power cycle etc. At boot nRF52 update resumes and user sees blinking red, assuming broken device.

Let's change the start self-update period to 2 hours so users are more likely to finish the configuration before update starts.

TheSomeMan:


ojousima (approved):
Thanks

PR #1255, opened 2026-04-10
Close #1254: Gateway sometimes reboots due to software assertion when connecting to Wi-Fi network

### LINKED ISSUE #1254: Gateway sometimes reboots due to software assertion when connecting to Wi-Fi network

TheSomeMan:


ojousima (approved):
Thanks

Issue #1261, opened 2026-05-05
nvs_flash: Add support for partial reading


TheSomeMan:

TheSomeMan:
@markoaamunkajo this task is a subtask of #1052, #1264, so the testing procedure is the same - just try to configure Gateway to send data to HTTP target with path/query up to 8KiB, and try to configure additional HTTP headers up to 8KiB, check on the server side that path/query/headers are present in HTTP request and they equal to expected values.

markoaamunkajo:
QA passed

PR #1262, opened 2026-05-05
[#1261] Copy components/nvs_flash from ESP-IDF v4.2.5

TheSomeMan:
This PR contains only copies of "nvs_flash" source files from ESP-IDF v4.2.5 without any modifications.


ojousima (approved):
Thanks

PR #1285, opened 2026-06-02
Fix #1284: ci: Firmware built in GitHub Actions embeds a commit hash instead of the "git describe tag"

### LINKED ISSUE #1284: Firmware built in GitHub Actions embeds a commit hash instead of the `git describe` tag
# Firmware built in GitHub Actions embeds a commit hash instead of the `git describe` tag

## Summary

When the firmware is built locally, the embedded application version is derived
from the most recent reachable git tag, e.g. `v1.17.1-48-g58af0ca8` (or
`v1.17.1-45-gc5123833-dirty` for a dirty working tree).

When the same commit is built by the `Build Firmware (dev)` /
`Build Firmware (prod)` GitHub Actions workflows, the embedded version is the
short commit hash only, e.g. `58af0ca`. The tag-based prefix is missing.

This breaks the version semantics shown in the Web UI, MQTT/HTTP payloads,
OTA metadata and release artifacts: there is no way to tell which release the
CI-built firmware is based on without manually mapping the hash to a tag.

## Affected workflows

- `.github/workflows/build-fw-dev.yml`
- `.github/workflows/build-fw-prod.yml`

## Reproduction

1. Push any commit (not directly on a tag) to `master`.
2. Wait for `Build Firmware (dev)` to finish.
3. Inspect the build log – look for the `Project version:` and
   `App "ruuvi_gateway_esp" version:` lines.
4. Observe the value is the 7-character commit hash instead of
   `vX.Y.Z-<n>-g<hash>`.

### Evidence from the failing run

Source: `logs_71932047322.zip` → `Build Firmware (dev)/2_Check out code.txt`:

```text
fetch-depth: 1
fetch-tags: false
[command]/usr/bin/git -c protocol.version=2 fetch --no-tags --prune \
    --no-recurse-submodules --depth=1 origin \
    +58af0ca8dc67feaaf697195b7d1bd7c1f40fd63a:refs/remotes/origin/master
 * [new ref]         58af0ca8dc67feaaf697195b7d1bd7c1f40fd63a -> origin/master
[command]/usr/bin/git checkout --progress --force -B master refs/remotes/origin/master
```

Source: `Build Firmware (dev)/11_Run build.txt`:

```text
Project version: 58af0ca
-- App "ruuvi_gateway_esp" version: 58af0ca
```

For the same commit, locally:

```text
$ git describe --tags --dirty --always
v1.17.1-48-g58af0ca8
```

## Root cause

The root `CMakeLists.txt` derives the embedded firmware version using
`git describe`:

```cmake
execute_process (
        COMMAND bash -c "git describe --always --tags --dirty"
        OUTPUT_VARIABLE PROJECT_VER
        OUTPUT_STRIP_TRAILING_WHITESPACE
)
message("Project version: ${PROJECT_VER}")
```

This value is later picked up by ESP-IDF's `app_update` component and written
into the `esp_app_desc` structure embedded in the firmware image.

`git describe --tags` needs two things to produce `vX.Y.Z-<n>-g<hash>`:

1. The most recent reachable annotated/lightweight tag must exist in the local
   repository.
2. The commit history between that tag and `HEAD` must be present so the
   commit-count `<n>` can be computed.

The CI checkout step (`actions/checkout@v6`) uses its defaults:

- `fetch-depth: 1` → only `HEAD` is fetched, no ancestors.
- `fetch-tags: false` → no tags are fetched.

Therefore `git describe --tags` finds no reachable tag, and because the script
passes `--always`, git falls back to the abbreviated commit hash – exactly
what we see in the build logs.

### Reproduction outside CI

Cloning the repository the same way GitHub Actions does it:

```bash
git clone --depth=1 --no-tags "file:///home/alexey/dev/ruuvi.gateway_esp.c" shallow
cd shallow
git rev-parse --is-shallow-repository   # -> true
git tag -l | wc -l                      # -> 0
git describe --always --tags --dirty    # -> 58af0ca   (CI behaviour)
```

After restoring full history and tags:

```bash
git fetch --prune --unshallow --tags
git rev-parse --is-shallow-repository   # -> false
git tag -l | wc -l                      # -> 51
git describe --always --tags --dirty    # -> v1.17.1-48-g58af0ca  (expected)
```

This confirms the cause is purely the shallow, tag-less CI checkout, not a
difference in CMake, ESP-IDF, the build environment or `git` itself.

## Fix

Configure `actions/checkout` in both firmware workflows to fetch the full
history and all tags:

```yaml
- name: Check out code
  uses: actions/checkout@v6
  with:
    fetch-depth: 0
    fetch-tags: true
    submodules: recursive
```

- `fetch-depth: 0` brings in all commits, so `git describe` can walk back to
  the latest reachable tag and compute the commit distance.
- `fetch-tags: true` makes the intent explicit and stays correct even if the
  default for `fetch-depth: 0` changes in future versions of
  `actions/checkout`.

## Files to change

- `.github/workflows/build-fw-dev.yml`
- `.github/workflows/build-fw-prod.yml`

The `SonarCloud Analysis` workflow already uses `fetch-depth: 0`, so it is not
affected. Other workflows (clang-format, schemas, unit tests, mbedtls, nvs)
do not depend on the firmware version string and do not need changes.

## Verification

After applying the fix, the CI log should contain:

```text
Project version: v1.17.1-<n>-g<hash>
-- App "ruuvi_gateway_esp" version: v1.17.1-<n>-g<hash>
```

and the same string should appear in the Web UI's "About" page and in the
`esp_app_desc.version` field of the generated `ruuvi_gateway_esp.bin`.

TheSomeMan:
## Summary

The `Build Firmware (dev)` and `Build Firmware (prod)` GitHub Actions
workflows produced firmware whose embedded `esp_app_desc.version` was just
the short commit hash (e.g. `58af0ca`) instead of the expected
tag-relative version produced by local builds
(e.g. `v1.17.1-48-g58af0ca8`).

This PR configures `actions/checkout` in both firmware workflows to fetch
the full git history and all tags, so `git describe --always --tags --dirty`
(used by the root `CMakeLists.txt` to populate `PROJECT_VER`) can resolve
the nearest reachable tag and compute the commit distance, matching the
behavior of local builds.

## Motivation

The embedded firmware version is surfaced in:

- the Web UI "About" page,
- HTTP/MQTT status payloads,
- OTA metadata (`esp_app_desc.version`),
- release artifacts and their filenames.

Showing only a 7-character commit hash makes it impossible to tell which
release a CI-built firmware is based on without manually mapping the hash
to a tag. It also breaks any downstream tooling that parses the
`vX.Y.Z-<n>-g<hash>` format.

## Root cause

`actions/checkout@v6` defaults to a shallow, tag-less checkout:

```text
fetch-depth: 1
fetch-tags: false
```

The build then runs:

```cmake
execute_process(
        COMMAND bash -c "git describe --always --tags --dirty"
        OUTPUT_VARIABLE PROJECT_VER
        OUTPUT_STRIP_TRAILING_WHITESPACE)
```

With no tags fetched and no ancestor commits available, `git describe
--tags` finds no reachable tag and, because of `--always`, falls back to
the abbreviated commit hash.

Full analysis, log excerpts and a local reproduction are documented in
[`docs/issues/ci-firmware-version-uses-commit-hash-instead-of-git-tag.md`](../docs/issues/ci-firmware-version-uses-commit-hash-instead-of-git-tag.md).

## Changes

- `.github/workflows/build-fw-dev.yml`: add `fetch-depth: 0` and
  `fetch-tags: true` to the `Check out code` step.
- `.github/workflows/build-fw-prod.yml`: same change.
- `docs/issues/ci-firmware-version-uses-commit-hash-instead-of-git-tag.md`:
  new file describing the bug, root cause and fix.

Diff applied to both firmware workflows:

```diff
       - name: Check out code
         uses: actions/checkout@v6
         with:
+          fetch-depth: 0
+          fetch-tags: true
           submodules: recursive
```

### Why both options

- `fetch-depth: 0` brings in the full commit history so `git describe`
  can walk back to the most recent reachable tag and compute the
  commit-count `<n>`.
- `fetch-tags: true` makes the intent explicit and remains correct even
  if the default behavior of `actions/checkout` regarding tags changes
  in a future major version.

## Scope / non-goals

- Only the two firmware build workflows are changed. The
  `SonarCloud Analysis` workflow already uses `fetch-depth: 0`, and the
  other workflows (clang-format, JSON schemas, Google tests, mbedTLS,
  nvs_flash) do not depend on the firmware version string.
- No source code, `CMakeLists.txt`, `sdkconfig` or build flag is changed.
  The fix is purely a CI configuration change.

## How to verify

1. Trigger `Build Firmware (dev)` for any commit that is not directly
   on a tag.
2. In the `Run build` step look for these lines:

    ```text
    Project version: v1.17.1-<n>-g<hash>
    -- App "ruuvi_gateway_esp" version: v1.17.1-<n>-g<hash>
    ```

    Before the fix they read just the short hash, e.g. `58af0ca`.

3. Flash the resulting `ruuvi_gateway_esp.bin` (or read it back with
   `esptool.py`) and confirm `esp_app_desc.version` matches the
   `git describe` output shown in the build log and in the Web UI's
   "About" page.

### Local reproduction of the original failure (no CI required)

```bash
git clone --depth=1 --no-tags https://github.com/ruuvi/ruuvi.gateway_esp.c shallow
cd shallow
git describe --always --tags --dirty   # -> short commit hash (buggy)
git fetch --prune --unshallow --tags
git describe --always --tags --dirty   # -> vX.Y.Z-<n>-g<hash> (fixed)
```

## Risk and rollback

- Risk: minimal. The change only increases the amount of git data
  fetched during CI. Build time grows by a few seconds for the
  additional history and tags.
- Rollback: revert the two-line addition in each workflow file.

### DIFF
--- a/.github/workflows/build-fw-dev.yml
+++ b/.github/workflows/build-fw-dev.yml
@@ -39,6 +39,8 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
         with:
+          fetch-depth: 0
+          fetch-tags: true
           submodules: recursive
 
       - name: Set up Python 3.8
--- a/.github/workflows/build-fw-prod.yml
+++ b/.github/workflows/build-fw-prod.yml
@@ -36,6 +36,8 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
         with:
+          fetch-depth: 0
+          fetch-tags: true
           submodules: recursive
 
       - name: Set up Python 3.8


ojousima (approved):
Thanks

Issue #1163, opened 2026-02-25
When connection is lost during the firmware update download, incorrect error message is displayed


TheSomeMan:

TheSomeMan:
@markoaamunkajo here you need to check UI after firmware download error (you can disconnect Ethernet cable to simulate error)

markoaamunkajo:
QA passed

PR #1291, opened 2026-06-03
Fix #1172: Memory leak on the failure path of fw_update_ota_partition

### LINKED ISSUE #1172: Memory leak on the failure path of fw_update_ota_partition
https://github.com/ruuvi/ruuvi.gateway_esp.c/pull/1171#pullrequestreview-3870543887

In the failure path of fw_update_ota_partition(...), the function returns without closing/cleaning up out_handle. With esp_ota_begin_patched() allocating an ota_ops_entry_t that is only freed in esp_ota_end_patched(), this early return will leak the OTA ops entry (and keep it in the global list), potentially exhausting memory over repeated attempts. Ensure the handle is finalized/aborted on this path (e.g., call esp_ota_end_patched(out_handle, ...) for cleanup and ignore its return, or add/use an esp_ota_abort_patched() equivalent).

TheSomeMan:
# Fix #1172: Memory leak on the failure path of fw_update_ota_partition
Related issue: #1163 

## Problem

In `fw_update_ota()`, after `esp_ota_begin_patched()` successfully allocates an
`ota_ops_entry_t` and inserts it into the global OTA ops linked list, a
subsequent failure in `fw_update_ota_partition()` (firmware download/write error)
causes an early return **without** calling `esp_ota_end_patched()` to
finalize and free the OTA handle.

This leaks the `ota_ops_entry_t` allocation and leaves a stale entry in the
global list. Over repeated failed update attempts (e.g. unstable network), this
can exhaust memory.

## Root Cause

`esp_ota_begin_patched()` allocates and list-inserts the OTA ops entry.
`esp_ota_end_patched()` is the only function that removes and frees it.
The failure path after a failed `fw_update_ota_partition()` call was missing the
`esp_ota_end_patched()` cleanup — it logged the error, set the error message,
and returned `false` immediately.

## Fix

Call `esp_ota_end_patched(out_handle, &pub_key_digest)` on the failure path
before returning. The return value is intentionally ignored (cast to `void`)
because this is a cleanup/abort path — `esp_ota_end_patched()` will report
validation errors (incomplete image), but it **unconditionally** removes the
entry from the global list and frees its memory regardless of the return code.

### Changed file

- **`main/fw_update.c`** — added 3 lines: declare a dummy `pub_key_digest`,
  call `esp_ota_end_patched()` to release the OTA handle on the download
  failure path.

## Verification

- Firmware builds successfully with `idf.py build`.
- Code reviewed against `esp_ota_end_patched()` implementation
  (`main/esp_ota_ops_patched.c`): confirmed that `LIST_REMOVE()` + `os_free()`
  execute unconditionally after the handle is found, regardless of
  validation/write errors.

### DIFF
--- a/main/fw_update.c
+++ b/main/fw_update.c
@@ -1116,7 +1116,10 @@ fw_update_ota(const char* const p_url, fw_update_error_message_info_t* const p_e
     if (!fw_update_ota_partition(p_partition, out_handle, p_url))
     {
         LOG_ERR("Failed to download firmware image");
-        p_error_message_info->p_message = "Firmware image download ended with an error";
+        p_error_message_info->p_message        = "Firmware image download ended with an error";
+        esp_ota_sha256_digest_t pub_key_digest = { 0 };
+        // esp_ota_end_patched removes the OTA ops entry even if validation fails.
+        (void)esp_ota_end_patched(out_handle, &pub_key_digest);
         return false;
     }
 


ojousima (approved):
Thanks. This seems a bit dangerous in the sense that function relies on `esp_ota_begin_patched` not allocating the memory in case of non-OK error code being returned. 

This holds true in current implementation, but if implementation ever changes to a form where memory is allocated in error case this function will reintroduce a potential memory leak. 

I'd consider calling the _end in case of failed begin, even if all it does is to check for a null pointer and return without doing anything.

TheSomeMan:
Thanks.

> This seems a bit dangerous in the sense that function relies on `esp_ota_begin_patched` not allocating the memory in case of non-OK error code being returned.

I'd say it's common assumption, that if a function returns error, it should perform cleanup and free all partially allocated resources. It looks too overprotective if callers will check and free resources after failed begin/init/open calls.

> This holds true in current implementation, but if implementation ever changes to a form where memory is allocated in error case this function will reintroduce a potential memory leak.

Since the implementation was already copied to our own repo, it will not change it case of ESP-IDF upgrade, so the risk is quite low.

> I'd consider calling the _end in case of failed begin, even if all it does is to check for a null pointer and return without doing anything.

I checked this out, and it looks like it can be done, but it looks more dangerous to me. `_begin` returns a handle (not pointer to allocated memory) and if '_begin' fails, the handle remains initial state (zero). But if variable initializer will be removed (which looks quite harmless), this will result in an uninitialized random value being used in an attempt to free the handle.

PR #1295, opened 2026-06-15
chore(fatfs): #1105 import components from ESP-IDF v4.2.2

TheSomeMan:
Copy components/fatfs/diskio/diskio_rawflash.c from ESP-IDF v4.2.2 to local components on branch v1.16.x.

Refs #1105


ojousima (approved):
Thanks

Issue #1294, opened 2026-06-14
Crash in `vQueueDelete(NULL)` when one FATFS mount call falls back to the other on the same partition


TheSomeMan:
## Summary

`flashfatfs_mount()` mounts a partition using `esp_vfs_fat_rawflash_mount()`
(wear-levelled FAT) and, on failure, falls back to
`esp_vfs_fat_spiflash_mount()` on the same partition so the firmware can
handle both image formats. When the first mount fails, the second mount
aborts the gateway on an assertion inside `vQueueDelete()`:

```
W (2596406) fatfs: find_volume: res=13
W (2596433) vfs_fat_spiflash: f_mount failed (13)
E (2596439) FlashFatFS: esp_vfs_fat_rawflash_mount failed, err=-1 (ESP_FAIL)
W (2596456) FlashFatFS: Try to mount partition as SPI-Flash FATFS
W (2596465) fatfs: Mounting logical drive 1
W (2596470) fatfs: cfs: 0x3ffeaaf8
W (2596474) fatfs: ff_del_syncobj 0x0
/home/.../components/freertos/queue.c:1816 (vQueueDelete)- assert failed!
```

The bug is **symmetric**: the same crash also occurs in the reverse
direction, i.e. when `esp_vfs_fat_spiflash_mount()` is attempted first,
fails, and is followed by `esp_vfs_fat_rawflash_mount()` on the same
partition. The order in which the two mount variants are called does not
matter — what matters is that any failed mount of one variant leaves
`FatFs[vol]` in a corrupted state for the next mount on the same volume.

## Affected code paths

- `main/flashfatfs.c` → `flashfatfs_mount()`, which tries
  `esp_vfs_fat_rawflash_mount()` first and falls back to
  `esp_vfs_fat_spiflash_mount()` so the same partition can be mounted as
  either raw-FAT or WL-FAT. The reverse fallback order has the same
  problem.
- ESP-IDF v4.2.x `components/fatfs/vfs/vfs_fat_spiflash.c` — root cause.
  Both `esp_vfs_fat_rawflash_mount()` and `esp_vfs_fat_spiflash_mount()`
  contain the same incomplete cleanup logic.

## Root cause

The `fail:` paths of both `esp_vfs_fat_rawflash_mount()` and
`esp_vfs_fat_spiflash_mount()` in ESP-IDF v4.2.x are incomplete:

[...]

PR #1299, opened 2026-06-15
chore(fatfs): #1294 import vfs_fat_spiflash from ESP-IDF v4.2.2

TheSomeMan:
Copy components/fatfs/diskio/vfs/vfs_fat_spiflash.c from ESP-IDF v4.2.2 to local components on branch v1.16.x.

Refs #1294


ojousima (approved):
Thanks

PR #1305, opened 2026-06-16
chore(fatfs): #1294 import vfs_fat_spiflash from ESP-IDF v4.2.5

TheSomeMan:
Copy components/fatfs/diskio/vfs/vfs_fat_spiflash.c from ESP-IDF v4.2.5 to local components on the master branch (version v1.17.x).

Refs #1294


ojousima (approved):
Thanks

PR #1308, opened 2026-06-16
ci(esp_crt_bundle): close #1306 detect outdated bundled Mozilla CA certificates

### LINKED ISSUE #1306: CI: detect when the embedded Mozilla CA bundle (esp_crt_bundle/cacert.pem) is out of date
# Issue description

## Summary

The firmware ships with a snapshot of the Mozilla CA root certificates
(`esp_crt_bundle/cacert.pem`, sourced from <https://curl.se/docs/caextract.html>).
Today there is nothing that warns us when curl publishes a newer bundle, so the
list silently drifts behind upstream — which over time means trusting roots
that have been removed by Mozilla and not trusting newly added ones.

## Proposal

Add a small piece of CI that compares the local bundle's date with the latest
one published by curl and fails loudly when the local copy is older.

### Acceptance criteria

- A helper script (e.g. `scripts/check_ca_bundle_up_to_date.sh`) that:
  - reads the `## Certificate data from Mozilla as of: …` header from
    `esp_crt_bundle/cacert.pem`,
  - downloads <https://curl.se/ca/cacert.pem>,
  - compares the two dates and exits non-zero when the remote is newer,
  - supports a `--update` flag to overwrite the local file in place.
- A GitHub Actions workflow that runs the script on every push / PR and on a
  weekly schedule.
- When the weekly scheduled run detects a stale bundle, a tracking issue is
  opened (and auto-closed once the bundle is refreshed) so the maintainer is
  notified without having to watch the Actions tab.

### Out of scope

- Automatically committing the new bundle. Updates should remain a deliberate
  human action so the diff can be reviewed.

TheSomeMan:
Closes #1306.

## What

- Add `scripts/check_ca_bundle_up_to_date.sh`, which compares the date embedded
  in `esp_crt_bundle/cacert.pem` (`## Certificate data from Mozilla as of: …`)
  with the latest bundle published at <https://curl.se/ca/cacert.pem>.
  Supports `--update` to overwrite the local file in place.
- Add `.github/workflows/check-ca-bundle.yml`, running the script on push, PR,
  weekly cron (Mon 06:00 UTC) and manual dispatch.
- On a **scheduled** run the workflow splits into a separate `notify` job that
  opens (or refreshes) a tracking issue when the bundle is stale and closes
  it on the next successful scheduled run. The issue title and body are
  defined inline in an `actions/github-script` step (no third-party action
  and no separate template file). The notifier uses the script's exit code
  (exposed as a job output) to distinguish "bundle is stale" (exit 1, toggle
  the issue) from internal errors such as download or parsing failures
  (exit 2, leave the issue untouched) so transient outages don't produce
  misleading alerts.
- The `notify` job is the only place where `issues: write` is granted, and it
  only runs on `schedule` — push and PR runs receive a read-only token.
- Rename `esp_crt_bundle/cacert-2025-05-20.pem` → `esp_crt_bundle/cacert.pem`
  and update `CONFIG_MBEDTLS_CUSTOM_CERTIFICATE_BUNDLE_PATH` in `sdkconfig`
  accordingly. The date is now read from the file's header, so the filename
  no longer has to change with each update — `sdkconfig` stays stable.

## Why

There was no way to notice when the embedded Mozilla CA root certificate
bundle fell behind upstream. Over time this means trusting roots that have
been removed by Mozilla and missing newly-added ones.

## Build impact

`components/mbedtls/CMakeLists.txt` already lists the custom bundle as a
`DEPENDS` of the `gen_crt_bundle.py` custom command, so changing
`cacert.pem` automatically regenerates the embedded `x509_crt_bundle` blob and
relinks the firmware. No additional build wiring is needed.

## How to test

```bash
# Should currently report the bundle as stale (exit 1):
bash scripts/check_ca_bundle_up_to_date.sh

# Update in place (exit 0):
bash scripts/check_ca_bundle_up_to_date.sh --update

# Re-run — now reports up-to-date (exit 0):
bash scripts/check_ca_bundle_up_to_date.sh
```

In CI, the new **Check CA Bundle** workflow runs automatically.

## Notes for reviewers

- The `notify` job needs `issues: write` permission to open/close the tracking
  issue; this is granted with a per-job `permissions:` block on that job
  only, so push/PR runs of `check-ca-bundle` use a read-only token. No
  repo-wide setting change is required as long as
  *Settings → Actions → General → Workflow permissions* allows per-workflow
  write scopes (the default).
- Scheduled workflows only run from the default branch, so the tracking-issue
  alerting will start working once this is merged to `master`.
- The tracking issue is identified purely by its (constant) title; no
  dedicated labels are added because the repository does not define any.


ojousima (approved):
Thanks.

Issue #1309, opened 2026-06-17
Sign v1.16.x release firmware while keeping unsigned-image OTA compatibility


TheSomeMan:
# Bridge release: signed v1.16.3 to enable smooth migration between unsigned v1.16.x and signed v1.17.x

## Title
`Release v1.16.3 as a transitional bridge: build signed firmware (Secure Boot v2) that still accepts unsigned OTA images, enabling smooth migration to v1.17.x and rollback to v1.16.2`
## Background
The Ruuvi Gateway firmware is split into two release lines:
* **v1.16.x and earlier** — released as **unsigned** firmware. All
  units currently in the field run images that carry no Secure Boot
  signature. Their OTA logic does not perform any signature check.
* **v1.17.x (next major)** — released as **signed** firmware. From
  v1.17.0 onwards every image carries a Secure Boot v2 signature and
  the OTA logic enforces signature verification.
Without a transitional release, devices running v1.16.2 (unsigned)
cannot be safely moved to v1.17.x (signed): a direct upgrade leaves
no working downgrade path, and any future need to roll back to a
v1.16.x image would fail because v1.17.x rejects unsigned payloads.
**v1.16.3 is therefore an intermediate / bridge release.** Its sole
purpose is to make the transition between the unsigned v1.16.x line
and the signed v1.17.x line smooth and reversible:
* it is itself **signed**, so it is byte-for-byte compatible with the
  v1.17.x signing scheme and can be installed by any v1.17.x device;
* it **also accepts unsigned OTA images**, so it can still be reached
  from v1.16.2 (unsigned) and can fall back to v1.16.2 if needed.
In other words, v1.16.3 is the only release that knows how to talk to
**both** worlds.
```
   unsigned                    bridge                        signed
 +----------+  OTA upgrade   +----------+  OTA upgrade   +----------+
 | v1.16.2  | -------------> | v1.16.3  | -------------> | v1.17.x  |
 |(unsigned)| <------------- | (signed) |                | (signed) |
 +----------+   rollback     +----------+                +----------+
                                  ^                           |
                                  +------ rollback -----------+
```
To achieve that, v1.16.3 ships with Secure Boot v2 signing **enabled
at build time** (so its binaries are signed), but Secure Boot is
deliberately **not burned into eFuse** (`CONFIG_SECURE_BOOT_INSECURE=y`),
and the application-side runtime signature check on OTA is suppressed
via a small downstream patch to ESP-IDF.
[...]

Issue #1102, opened 2025-12-22
Compare pub_key digest of downloaded firmware with pub_key digest of currently running firmware


TheSomeMan:
If Secure Boot is not active (old bootloader is used), then eFuse does not contain pub-key digest and signature validation always returns 'valid' for any signed firmware. 
Need to compare the pub-key digest of downloaded firmware with the pub-key digest of currently running firmware

TheSomeMan:
@markoaamunkajo to test this use the same steps as in #1094. Check in logs that the firmware signature is checked before rebooting.

markoaamunkajo:
Downgrade to v1.16.x, correct
Update to alpha, correct
QA passed

Issue #1094, opened 2025-12-09
Generate signed firmware images, verify signature on update


TheSomeMan:

TheSomeMan:
@markoaamunkajo to test this, you need to install firmware from the master branch and try to downgrade to v1.16.x or update to firmware from development branch. In both cases firmware signature verification should fail and original firmware should stay intact.

markoaamunkajo:
Downgrade to v1.16.x, correct
Update to alpha, correct
QA passed

markoaamunkajo:
Downgrade to 1.16x, correct
Update to alpha, correct

PR #1314, opened 2026-06-18
chore(bootloader_support): #1309 vendor patched bootloader_support from ESP-IDF v4.2.2

TheSomeMan:
## Summary

Back-port the patched `components/bootloader_support` from the `v1.17.x`
branch of `ruuvi.gateway_esp.c` into `v1.16.x`. The component is a fork
of the ESP-IDF v4.2.2 `components/bootloader_support` with a minimal,
targeted set of changes required by the Ruuvi Gateway firmware:

* enable signature verification of OTA images from non-bootloader
  (application) code,
* allow inclusion of `esp_app_format.h` from C++ translation units,
* avoid a stack overflow during RSA-PSS signature verification of OTA
  images, and
* stay source-compatible with Mbed TLS 3.x in addition to the 2.x API
  that ships with ESP-IDF v4.2.2.

The component is otherwise a verbatim copy of the upstream ESP-IDF
v4.2.2 sources, so the diff against upstream is intentionally small and
easy to audit.

## Motivation

Ruuvi Gateway verifies the digital signature of OTA images from
application code (see #1094 / #1103). The stock IDF v4.2.2
`bootloader_support` component:

1. Exposes `esp_secure_boot_verify_rsa_signature_block()` only through
   the *private* `include_bootloader` directory, which is not visible
   outside the bootloader build.
2. Allocates the entire RSA verification working set (key digests,
   `mbedtls_rsa_context`, entropy and DRBG contexts) on the stack —
   roughly several hundred bytes — which overflows the stack of the
   task that performs OTA verification on the gateway.
3. Uses the Mbed TLS 2.x API only, which makes it impossible to build
   against newer Mbed TLS releases.
4. Declares `esp_app_format.h` without `extern "C"` guards and uses
   `_Static_assert`, which prevents inclusion from C++ sources.

Forking the component is the least invasive way to address all four
issues without modifying the in-tree ESP-IDF and without disabling the
upstream's CI / signing tooling.

## Changes vs. ESP-IDF v4.2.2 baseline

Diff is limited to three files:

### `components/bootloader_support/CMakeLists.txt`

* `include` *and* `include_bootloader` are now registered as **public**
  include directories regardless of whether the component is built as
  part of the bootloader or the application. Upstream marks
  `include_bootloader` as `priv_include_dirs` for the application
  build, which hides
  `esp_secure_boot_verify_rsa_signature_block()` from OTA code.

### `components/bootloader_support/include/esp_app_format.h`

* Added `extern "C" { ... }` guards.
* Wrapped the three `_Static_assert` declarations in
  `#ifndef __cplusplus` so the header can be included from C++
  translation units (where `_Static_assert` is not a keyword in the
  same way and the assertions are not needed for the C++ side).

### `components/bootloader_support/src/idf/secure_boot_signatures.c`

* Introduced `verify_rsa_sig_block_mem_t`, a single structure that
  groups all sizeable working buffers used during signature
  verification:
  * per-block SHA-256 digests of the signature-block public keys,
  * the trusted digest read back from eFuse,
  * the all-zero comparison buffer,
  * the `mbedtls_rsa_context`,
  * the `mbedtls_entropy_context`, and
  * the `mbedtls_ctr_drbg_context`.
* The existing logic was refactored into a static
  `esp_secure_boot_verify_rsa_signature_block_internal()` worker that
  takes a pointer to this structure.
* The public
  `esp_secure_boot_verify_rsa_signature_block()` now `calloc()`s the
  structure, delegates to the worker, and `free()`s the memory on
  return. This moves the verification working set from the caller's
  stack to the heap and prevents stack overflows in tasks with small
  stacks (e.g. the OTA / HTTPS worker tasks).
* Added Mbed TLS 3.x compatibility, gated on
  `MBEDTLS_VERSION_MAJOR >= 3`:
  * `mbedtls_rsa_init()` is called with the new single-argument
    signature and padding is configured separately via
    `mbedtls_rsa_set_padding()`.
  * `mbedtls_rsa_rsassa_pss_verify()` is called with the
    Mbed TLS 3.x signature (no RNG callback, no `MBEDTLS_RSA_PUBLIC`
    mode argument).
* Added the necessary `#include "mbedtls/rsa.h"` and
  `#include "mbedtls/version.h"` for the version detection above.

No other files were modified. The full list of changed files can be
reproduced with:

```sh
diff -rqN \
    $IDF_PATH/components/bootloader_support \
    components/bootloader_support
```

against an unmodified ESP-IDF v4.2.2 checkout.

## Risk / impact

* **Bootloader image is unchanged.** All edits are either build-system
  scope (header visibility), C++-only guards, or in the IDF-side
  `src/idf/secure_boot_signatures.c` which is only compiled into the
  application, not the bootloader.
* **Secure-boot algorithm is unchanged.** The cryptographic operations
  performed (SHA-256 of the public keys, eFuse comparison, RSA-PSS
  verification) are bit-for-bit identical to upstream; only the
  storage of intermediate values moves from the stack to the heap.
* **Mbed TLS 3.x branches are dead code** when building against the
  Mbed TLS 2.x that ships with ESP-IDF v4.2.2, so behaviour on the
  shipping toolchain is unaffected.

## Testing

* Builds with `idf.py build` against ESP-IDF v4.2.2 (the toolchain
  used by `v1.16.x`).
* Signed OTA flow verified end-to-end on hardware (signature accepted
  for a correctly signed image, rejected for a tampered image, no
  stack-overflow panic on the verifying task).
* Diff against upstream ESP-IDF v4.2.2 contains only the three files
  listed above.

## Related

* Issue: #1309
* Source branch: `v1.17.x` (the patched component is copied verbatim
  from there).
* Earlier related work: #1094 / #1095 (signed OTA), #1102 / #1103
  (public-key digest comparison on OTA).

### DIFF
--- a/components/bootloader_support/CMakeLists.txt
+++ b/components/bootloader_support/CMakeLists.txt
@@ -1,0 +1,112 @@
+set(srcs
+    "src/bootloader_clock.c"
+    "src/bootloader_common.c"
+    "src/bootloader_flash.c"
+    "src/bootloader_mem.c"
+    "src/bootloader_random.c"
+    "src/bootloader_random_${IDF_TARGET}.c"
+    "src/bootloader_utility.c"
+    "src/esp_image_format.c"
+    "src/flash_encrypt.c"
+    "src/flash_partitions.c"
+    "src/flash_qio_mode.c"
+    "src/bootloader_flash_config_${IDF_TARGET}.c"
+    "src/bootloader_efuse_${IDF_TARGET}.c"
+    )
+
+set(include_dirs "include" "include_bootloader")
+if(BOOTLOADER_BUILD)
+    set(priv_requires micro-ecc spi_flash efuse)
+    list(APPEND srcs
+    "src/bootloader_init.c"
+    "src/${IDF_TARGET}/bootloader_sha.c"
+    "src/${IDF_TARGET}/flash_encrypt.c"
+    "src/${IDF_TARGET}/bootloader_${IDF_TARGET}.c"
+    )
+else()
+    list(APPEND srcs
+        "src/idf/bootloader_sha.c")
+    set(priv_requires spi_flash mbedtls efuse)
+endif()
+
+if(CONFIG_SECURE_SIGNED_APPS_ECDSA_SCHEME OR CONFIG_SECURE_SIGNED_APPS_RSA_SCHEME)
+    if(BOOTLOADER_BUILD)
+        list(APPEND srcs
+        "src/${IDF_TARGET}/secure_boot_signatures.c")
+    else()
+        list(APPEND srcs
+            "src/idf/secure_boot_signatures.c")
+    endif()
+endif()
+
+if(CONFIG_SECURE_BOOT AND BOOTLOADER_BUILD)
+    list(APPEND srcs
+        "src/${IDF_TARGET}/secure_boot.c")
+endif()
+
+set(requires soc) #unfortunately the header directly uses SOC registers
+
+idf_component_register(SRCS "${srcs}"
+                    INCLUDE_DIRS "${include_dirs}"
+                    PRIV_INCLUDE_DIRS "${priv_include_dirs}"
+                    REQUIRES "${requires}"
+                    PRIV_REQUIRES "${priv_requires}")
+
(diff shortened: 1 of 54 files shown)


ojousima (approved):
Thanks

Issue #1195, opened 2026-03-12
CRA: 303 645 IXIT-6-SOFTCOMP


ojousima:
## IXIT 6-SoftComp: Software Components

The completed IXIT lists all software components of the DUT. The pro forma contains the following entries and is typically filled out in form of a table.

> **NOTE 18:** The used level of detail concerning the division of the DUT software into software components serves for the fact that the TL can identify which components are updatable and which are not.

| Field | Description |
|-------|-------------|
| **ID** | Unique per IXIT identifier, that may be assigned using a sequential numbering scheme or some other labelling scheme. |
| **Description** | Brief description of the software component. |
| **Update Mechanism** | Reference to update mechanisms in IXIT 7-UpdMech that are used for updating the software component. An empty list of update mechanisms indicates the absence of updates for the software component and in this case a justification is provided. |
| **Cryptographic Usage** | Indicates if the software component makes use of cryptographic algorithms or primitives (Yes/No) and if so, whether side effects of replacing those algorithms and primitives are considered by the manufacturer (Yes/No). |

> **EXAMPLE 9 (ID):** Sequential numbering (`SoftComp-1`) or labelling scheme (`SoftComp-Firmw`).

> **NOTE 19 (Description):** BIOS, firmware and boot loader are possible software components of the DUT.

---
*Source: ETSI TS 103 701 V2.1.1 (2025-05)*

PR #1335, opened 2026-07-14
docs(cra) #1195 IXIT_6-SoftComp: update cryptographic usage description

TheSomeMan:


ojousima (commented):
SoftComp-nRF52FW: Did the nRF52811 have a boot time RAM calculation of SHA checksums of the flash contents? If yes, the cryptographic usage should be described briefly

ojousima (approved):
Thanks

PR #1348, opened 2026-07-15
docs(cra) #1347 IXIT_25-DelFunc: Deletion Functionalities

TheSomeMan:


ojousima (changes requested):
`ID: DelFunc-WebUI-Token-Erasure` 

Is this guaranteed to actually erase the data, or can data remain in wearleveling page of the NVS after saving new value? 

`DelFunc-Cloud-Relay-Deactivation` 

This section is covered by isolation, stopping relaying data is not deleting already existing data. 

The core purpose of data deletion is to describe to users how they can securely delete all their data from the device and the associated services, e.g. for secure disposal or resale of the hardware. 

In this case Id say less is more. If we give users only one path to delete their data from device (factory reset) and associated service (account deletion) we cover the requirements and minimise chances of misunderstanding. 

Draft for Cloud data deletion, feel free to adjust as necessary. 

```
ID: DelFunc-Service-Data-Erasure

Users can delete their data in associated service by signing in and requesting account deletion. Upon request, a deletion link is emailed to the user for confirmation of deletion. After user has followed the link, all of personally identifiable data is deleted and raw data, such as background pictures and raw sensor values, are queued for a background deletion process. 
```

ojousima (approved):
Thanks

Issue #1197, opened 2026-03-12
CRA: 303 645 IXIT-8-UpdProc


ojousima:
## IXIT 8-UpdProc: Update Procedures

This completed IXIT lists procedures of the manufacturer for the management of security updates. The pro forma contains the following entries and is typically filled out in form of a table.

| Field | Description |
|-------|-------------|
| **ID** | Unique per IXIT identifier, that may be assigned using a sequential numbering scheme or some other labelling scheme. |
| **Description** | Brief description of the procedure for deploying security updates including all entities and responsibilities. |
| **Time Frame** | Targeted time frame for completing the procedure. |

> **EXAMPLE 15 (ID):** Sequential numbering (`UpdProc-1`) or labelling scheme (`UpdProc-SecUpd`).

---
*Source: ETSI TS 103 701 V2.1.1 (2025-05)*

PR #1354, opened 2026-07-15
docs(cra) #1197 IXIT_8-UpdProc: update docs

TheSomeMan:


ojousima (approved):
I think 
```
4. Signing: The final firmware binaries are cryptographically signed using the manufacturer’s private RSA-3072 key (enforcing ESP32 Secure Boot v2 via RSA-PSS configuration parameters) within a hardware-isolated, secure build environment.
```
needs adjustment, I would understand hardware-isolated environment as completely offline PC which would sign the package without the signing key ever ending up on any Cloud infrastructure. In our case we use GitHub secrets, scoped to access-controlled branches. 

Otherwise good to go, thanks.

PR #1356, opened 2026-07-15
docs(cra) #1355 IXIT_28-LogIntf: Logical Interfaces

TheSomeMan:


ojousima (approved):
LogIntf-HTTP-Server exposes version information to unauthenticated user, 

Image

This should be documented, otherwise all good

TheSomeMan:
> LogIntf-HTTP-Server exposes version information to unauthenticated user,

Do we need to show this info to unauthenticated user? Maybe it would be better to disable it?

ojousima:
>Do we need to show this info to unauthenticated user? Maybe it would be better to disable it?

Yeah, that would be best. Opened #1380, let's implement it for v1.18.

Issue #1217, opened 2026-03-23
CRA: 303 645 IXIT-14-SecMgmt


ojousima:
## IXIT 14-SecMgmt: Secure Management Processes

The completed IXIT lists all secure management processes for critical security parameters implemented by the SO for the DUT. The pro forma contains the following entries and is typically filled out in form of a table.

| Field | Description |
|-------|-------------|
| **ID** | Unique per IXIT identifier, assigned using a sequential numbering scheme or other labelling scheme. _Example: `SecMgmt-1` or `SecMgmt-Passwd`_ |
| **Description** | Brief description of the secure management process regarding the whole life cycle for critical security parameters. If an existing standard is used, a reference to the corresponding standard is provided. The life cycle typically considers: generation, provisioning, storage, updates, decommissioning, archival, destruction, and processes to handle expiration and compromise of the parameter. |

PR #1361, opened 2026-07-15
docs(cra) #1217 IXIT_14-SecMgmt: Secure Management Processes

TheSomeMan:


ojousima (approved):
Hardware-Silicon-Root: I think there is no enforced change of the administrative password? 

This risk has been accepted on the basis that attacker could factory reset the Gateway and use default password in any case and sensitive information, such as passwords, are not exposed over the configuration interface? 

Regarding the HMAC-Signing key, it would be a better practice to have unique key provisioned to the Gateway and not expose that alongside the access point password, but as far as I remember the requirement is only for unique passwords / critical security parameters per device, reusing same secret is allowed. 

The threat here is that attacker with physical access to device could clone the signing mechanism and spoof the Gateway data, but this sort of attack would have to be targeted against each individual Gateway.

Otherwise all good, thanks

TheSomeMan:
> Hardware-Silicon-Root: I think there is no enforced change of the administrative password?
> 
> This risk has been accepted on the basis that attacker could factory reset the Gateway and use default password in any case and sensitive information, such as passwords, are not exposed over the configuration interface?
> 
> Regarding the HMAC-Signing key, it would be a better practice to have unique key provisioned to the Gateway and not expose that alongside the access point password, but as far as I remember the requirement is only for unique passwords / critical security parameters per device, reusing same secret is allowed.
> 
> The threat here is that attacker with physical access to device could clone the signing mechanism and spoof the Gateway data, but this sort of attack would have to be targeted against each individual Gateway.
> 
> Otherwise all good, thanks

You're correct that password changes are not forced during onboarding. Under ETSI EN 303 645 Provision 5.1-2, a pre-installed default password does not require a forced change if it is unique per device. Because our default password is derived from the 64-bit hardware-unique DEVICEID (burned into nRF52 FICR registers), it satisfies Provision 5.1-2 as a unique, high-entropy pre-installed credential. Custom password updates in the Web-UI remain supported but optional.

Sensitive Data Exposure: Confirmed - sensitive credentials (passwords, Wi-Fi keys, M2M tokens) are never exposed or returned in cleartext over configuration reads or Web-UI API responses.

HMAC Key & Physical Cloning Threat: You're right that using DEVICEID as the initial HMAC secret means an attacker with physical access could attempt to spoof telemetry for that specific gateway. However, because each DEVICEID is unique, this attack is strictly isolated to the single physical unit. Furthermore, Ruuvi Cloud can dynamically mitigate this by rotating the active HMAC key at runtime via the Ruuvi-HMAC-KEY response header over TLS. Therefore, if a second gateway attempts to send data with the initial signature while another active gateway is sending data with the rotated signature, that second gateway can be blocked in Ruuvi Cloud.

I updated `SecMgmt-Hardware-Silicon-Root` section and added `Compromise & Risk Management` subsection

Issue #1216, opened 2026-03-23
CRA: 303 645 IXIT-13-SoftServ


ojousima:
## IXIT 13-SoftServ: Software Services

The completed IXIT lists all software services of the DUT. The pro forma contains the following entries and is typically filled out in form of a table.

| Field | Description |
|-------|-------------|
| **ID** | Unique per IXIT identifier, assigned using a sequential numbering scheme or other labelling scheme. _Example: `SoftServ-1` or `SoftServ-WebServ`_ |
| **Description** | Brief description of the service, including its purpose. Indicates whether the service is accessible via a network interface and whether this is the case in the initialized state. _Example: An SSH daemon disabled by default because it was used only for development purposes._ |
| **Status** | Indication whether the service is **enabled** or **disabled** in the initialized state. |
| **Justification** | _(Enabled services only)_ Justification why the service is necessary for the intended use or operation of the DUT. |
| **Allows Configuration** | _(Network-accessible services only)_ Yes/No indication whether the service allows security-relevant changes in configuration, and if so, a brief description of the possible configuration. |
| **Authentication Mechanism** | _(Services allowing user access via network interface only)_ Reference to authentication mechanisms in IXIT 1-AuthMech used for authentication prior to use of the service. |

PR #1365, opened 2026-07-16
docs(cra) #1216 IXIT_13-SoftServ: Software Services

TheSomeMan:


ojousima (changes requested):
The "allows configuration" is inconsistent between `SoftServ-Telemetry-Relay-RuuviCloud` and `SoftServ-Telemetry-Relay-CustomHTTP`, Ruuvi Cloud is defined as "No" and Custom is defined as "Yes".

I'd lean towards No for both. These are outbound telemetry services,  the return-value of header `X-Ruuvi-Gateway-Rate` does allow adjusting the data transmission rate but that could be considered as not security relevant. as per ETSI TS 103 701 V2.1.1 IXIT pro forma A.3: `Indication whether the service allows security-relevant changes in configuration` Configuring the authentication for outbound connections is covered by `SoftServ-Local-Management-WebUI`, `SoftServ-Local-Programmatic-API` and `SoftServ-Centralized-Remote-Config`,

Similarly I'd consider `SoftServ-OTA-Firmware-Updater` as Allows configuration: No. While the service itself is configurable, the service does not allow configuration of device. This is also supported by description 

```
(update URL, auto-update cycle and schedule are user-configurable via the Web-UI), but the service does not expose any inbound configuration interface.
```

TheSomeMan:
> The "allows configuration" is inconsistent between `SoftServ-Telemetry-Relay-RuuviCloud` and `SoftServ-Telemetry-Relay-CustomHTTP`, Ruuvi Cloud is defined as "No" and Custom is defined as "Yes".

Yes, you're right, I misunderstood it a little. Good catch.
I've fixed it.

ojousima (approved):
Thanks

PR #1367, opened 2026-07-16
docs(cra) #1366 IXIT_21-PersData: Personal Data

TheSomeMan:


ojousima (changes requested):
I think we need to add sensor data collected as personal information. 

While sensor data itself is external to the Gateway, Gateway is clearly sold as a part of a system that collects sensor data. For example home occupancy can be determined from fridge temperature and shower humidity graphs, and the data becomes associated with user account, together with email which often contains name of the user.

ojousima (approved):
Thanks

PR #1384, opened 2026-07-30
docs(cra) #1237 ICS_5.3-1: All Non-Immutable Software Components are Securely Updateable

### LINKED ISSUE #1237: CRA: 303 645 ICS 5.3-1

TheSomeMan:
Close #1237


ojousima (changes requested):
The functional test claims
```
Functional checks confirm that all non-immutable
components ( SoftComp-SecondBoot, SoftComp-MainFW, SoftComp-nRF52FW,
SoftComp-WebUI) can be updated using their assigned mechanisms, and that
invalid, tampered, or improperly signed binaries cannot be executed under any
tested condition.
```

I think using the USB update mechanism would allow attacker to produce a binary with their own keys and print out the secrets in plain text over USB console. This should be documented, and then we need to decide if requirement of physical access is sufficient to fill the criteria or if we need to e.g. enforce encrypted storage and ensure that encryption keys are erased on USB update.

TheSomeMan:
> The functional test claims
> 
> ```
> Functional checks confirm that all non-immutable
> components ( SoftComp-SecondBoot, SoftComp-MainFW, SoftComp-nRF52FW,
> SoftComp-WebUI) can be updated using their assigned mechanisms, and that
> invalid, tampered, or improperly signed binaries cannot be executed under any
> tested condition.
> ```
> 
> I think using the USB update mechanism would allow attacker to produce a binary with their own keys and print out the secrets in plain text over USB console. This should be documented, and then we need to decide if requirement of physical access is sufficient to fill the criteria or if we need to e.g. enforce encrypted storage and ensure that encryption keys are erased on USB update.

Your concern regarding the physical attack over USB is technically accurate for the current v1.17.x architecture, but it should be clarified according to the regulatory context under ETSI EN 303 645 / ETSI TS 103 701:

1. Physical Access vs. NVS Readout:
In the current v1.17.x firmware release, NVS flash encryption is not enabled. Consequently, an attacker with physical access to the board does not actually need to build or flash a custom binary to exfiltrate NVS secrets — they can simply execute `esptool.py read_flash` over the USB-UART interface to dump the raw NVS partition (`nvs` / `gw_cfg_def`) directly.

2. ETSI EN 303 645 Compliance Scope (Provision 5.3-1):
Provision 5.3-1 ("Keep software updated") focuses primarily on ensuring that network-based software update mechanisms (UpdMech-WebUI and UpdMech-Auto) enforce cryptographic integrity and authenticity checks (RSA-3072-PSS / SHA-256) to prevent remote attackers from misusing the update channel (e.g., via MitM or malicious server impersonation). Under ETSI TS 103 701, local UART/ROM bootloader flashing (UpdMech-USB) is classified under physical interface boundaries (PhyIntf-USB), where physical proximity / enclosure access is the primary gating factor.

3. Updated Technical File & Roadmap (v1.18.x):
I have updated `IXIT 7-UpdMech` and `Self_Assessment_Test_Group_5_3_1.md` to:
- Explicitly clarify that `UpdMech-USB` requires local physical access to the USB-C interface (LogIntf-USB-Boot-Flasher / PhyIntf-USB) and operates outside the network stack.
- Precision-tune the functional assessment claims in Test Case 5.3-1-2 to distinguish between network update signature enforcement and local physical USB flashing.
- Note that hardware-enforced ESP32 Secure Boot v2 (burning public key hashes into eFuses) and NVS Flash Encryption are planned for the upcoming v1.18.x firmware release to harden flash memory against physical extraction vectors.

ojousima (approved):
Thanks

PR #1385, opened 2026-07-30
docs(cra) #1238 ICS_5.3-2: At Least One Secure Update Mechanism Exists

### LINKED ISSUE #1238: CRA: 303 645 ICS 5.3-2

TheSomeMan:
Close #1238


ojousima (commented):
I think this section needs discussion on 5.3-1 first, related to USB mechanism. 

The functional test covers official, valid v1.17.3 flashed alongside altered partitions, but would it be possible to downgrade to e.g. v1.16.2 over USB and then flash malicious partition?

ojousima (approved):
Thanks. 

The security and roadmap of USB updates is already discussed in #1384

PR #1415, opened 2026-07-30
docs(cra) #1414 ICS_5.4-1: Securely Store Sensitive Security Parameters

### LINKED ISSUE #1414: CRA: 303 645 ICS 5.4-1: Securely Store Sensitive Security Parameters

TheSomeMan:
Close #1414


ojousima (approved):
Thanks. 

These tests do detail protection against remote threats and correctly note that flash storage itself is unencrypted. The result is that security of user credentials is provided by the environment, e.g. locked door before Gateway device can be accessed. 

I would consider our current implementation to be compliant with the current regulation, even without at rest encryption, but it would be good to mention in our documentation that versions up to v.1.17.x are intended to be used only in areas which are not publicly accessible for security.

TheSomeMan:
> I would consider our current implementation to be compliant with the current regulation, even without at rest encryption, but it would be good to mention in our documentation that versions up to v.1.17.x are intended to be used only in areas which are not publicly accessible for security.

I updated the document and related IXIT_10 and IXIT_2 (in separate PRs)

PR #1441, opened 2026-07-31
docs(cra) #1440 ICS_5.6-2: Minimization of Disclosed Information in Unauthenticated Contexts

### LINKED ISSUE #1440: CRA: 303 645 ICS 5.6-2: Minimization of Disclosed Information in Unauthenticated Contexts

TheSomeMan:
Close #1440

### DIFF
--- a/cra/303645/Self_Assessment_Test_Group_5_6-2.md
+++ b/cra/303645/Self_Assessment_Test_Group_5_6-2.md
@@ -1,0 +1,103 @@
+# Test group 5.6-2: Minimization of Disclosed Information in Unauthenticated Contexts
+
+Provision 5.6-2 — Status: **M**. Related IXIT: `IXIT 28-LogIntf`.
+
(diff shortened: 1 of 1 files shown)


ojousima (changes requested):
~~I think I found a missing entry in IXIT-28-LogIntf: The /metrics endpoint is not documented.~~ 
~~Edit: The HTTP server is documented, documenting each individual URL path would be overkill.~~ 

~~So this seems more like the documented authentication is not being enforced on /metrics~~ 

~~This requires a bit of thought, enforcing authentication would break some existing installations. Would it be straightforward to put the /metrics behind access tokens?~~ 

~~I think I got confused here. Is the authentication to /metrics endpoint enforced?~~

ojousima (approved):
Thanks, I got initially confused over /metrics endpoint but these are covered by HTTP server.

PR #1519, opened 2026-08-13
fix(gw_cfg): #1518 Web UI displays “Custom Server” when “Ruuvi Cloud” is selected

### LINKED ISSUE #1518: Web-UI shows "Custom Server" even if default "Ruuvi Cloud" is selected (in v1.17.3)
Web UI displays “Custom Server” when “Ruuvi Cloud” is selected
In Gateway firmware v1.17.3, selecting the default Ruuvi Cloud HTTP destination may be displayed as Custom Server in the Web UI.
The generated gateway configuration should contain:
{
  "use_http_ruuvi": true,
  "use_http": true,
  "http_url": "https://network.ruuvi.com/record",
  "http_period": 10
}
Instead, stale HTTP values may be serialized, such as an empty http_url and an incorrect http_period, causing the Web UI to misidentify the selected destination.
Expected behavior: The Web UI displays Ruuvi Cloud and the configuration uses the canonical Ruuvi Cloud URL and 10-second reporting period.
Affected version: v1.17.3

TheSomeMan:
Close #1518
Use the initialized gateway HTTP defaults when serializing the Ruuvi HTTP configuration, ensuring the default URL and period are preserved instead of stale configuration values.

Add regression coverage for the expected Ruuvi HTTP settings.


ojousima (approved):
Thanks

PR #1520, opened 2026-08-15
fix(ui) #1380 Do not show firmware version info to unauthenticated user in Web-UI

### LINKED ISSUE #1380: Do not show firmware version info to unauthenticated user in Web-UI

TheSomeMan:
Close #1380


ojousima (approved):
Thanks

PR #1531, opened 2026-08-18
fix(http_server_cb): #1528 Fix query parameter boundary matching

### LINKED ISSUE #1528: Match URL query parameter names only at segment boundaries
## Problem

`http_server_get_from_params()` locates a key using unanchored `strstr()`.

Affected code: `main/http_server_cb.c`, currently around lines 705-725. `http_server_get_from_params_with_decoding()` inherits the behavior.

A requested key is therefore accepted when it appears:

- as a suffix of another parameter name, e.g. searching for `url=` in `noturl=https://example.com`;
- inside another parameter's value, e.g. searching for `url=` in `note=url=https://example.com`.

This allows malformed query strings to supply the wrong value to `validate_url()`. The same helpers are also used by file-related endpoints.

## Reproduction

These calls currently return a non-NULL value even though no `url` parameter exists:

```c
http_server_get_from_params("noturl=https://example.com", "url=");
http_server_get_from_params("note=url=https://example.com", "url=");
```

## Suggested fix

Parse/query parameter segments rather than accepting the first substring match. A key is valid only when it starts at the beginning of `p_params` or immediately after `&`. Continue searching if a substring match is not at a segment boundary.

Preserve current behavior for:

- parameters at the beginning, middle, and end;
- empty values;
- the existing first-match behavior for duplicate exact keys, unless deliberately specified otherwise.

## Regression coverage

Extend `tests/test_http_server_cb/test_http_server_cb.cpp` with direct tests for both helpers covering:

- a key that is a suffix of another parameter name;
- a key-like substring inside another value;
- a later valid exact key after an invalid substring match;
- normal and empty values.

## Expected result

Only exact query parameter names at segment boundaries are returned; malformed near-matches return NULL or are skipped in favor of a later exact key.

TheSomeMan:
- Match query parameter keys only at segment boundaries
- Skip suffix and value substring matches
- Preserve empty values and first valid match behavior
- Add regression tests for raw and decoded helpers

Fixes #1528

### DIFF
--- a/main/http_server_cb.c
+++ b/main/http_server_cb.c
@@ -16,7 +16,6 @@
 #include "http.h"
 #include "fw_update.h"
 #include "os_time.h"
-#include "time_str.h"
 #include "reset_task.h"
 #include "gw_cfg.h"
 #include "gw_cfg_json_parse.h"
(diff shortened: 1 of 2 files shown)


ojousima (approved):
Thanks,  the test cases make it easy to read through the effect of addition

PR #1532, opened 2026-08-18
fix(url_encode): #1529 reject percent-decoded NUL bytes

### LINKED ISSUE #1529: Reject percent-decoded NUL bytes in string URL parameters
## Problem

The URL decoder accepts `%00` and writes a NUL byte into a `str_buf_t` that is subsequently treated as a C string.

Affected code: `main/url_encode.c`, especially `url_n_decode_to_str_buf()` around lines 91-110. The behavior reaches request parsing through `http_server_get_from_params_with_decoding()` in `main/http_server_cb.c`.

For example, decoding `https%3A%2F%2Fgood.example%00ignored` creates storage containing a hidden suffix, but all downstream `strlen()`, logging, comparison, and URL parsing sees only `https://good.example`. The `str_buf_t.idx` can also disagree with the visible C-string length.

This is not the previously fixed incomplete-escape case: `%`, `%3`, and `%3Q` are rejected, while `%00` currently succeeds.

## Impact

Malformed request parameters can be silently truncated instead of rejected. This can create validation/interpretation differences and makes the string-returning decoder's result internally inconsistent.

## Suggested fix

Since these decoder APIs return C strings, reject an escape whose decoded byte is zero. Return failure through the existing `false`/NULL path so required `validate_url` parameters produce an error response.

Ensure any allocation made by `url_n_decode_with_alloc()` is freed if a decoding pass fails.

## Regression coverage

Add tests in `tests/test_url_encode/test_url_encode.cpp` for `%00` at the beginning, middle, and end of a value. Also add a helper or `/validate_url` regression test showing that a parameter containing `%00` is rejected rather than truncated.

Keep the existing successful decoding and malformed `%`, `%3`, and `%3Q` cases unchanged.

## Expected result

String URL-decoding APIs reject percent-encoded NUL bytes and never return a buffer whose logical content contains an embedded NUL.

TheSomeMan:
Reject decoded NUL bytes before writing them to C-string buffers. Free allocated decode buffers when decoding fails and add regression coverage for NUL bytes at every position and HTTP request parameters.

Fixes #1529

### DIFF
--- a/main/url_encode.c
+++ b/main/url_encode.c
@@ -106,6 +106,10 @@ url_n_decode_to_str_buf(const char* const p_src, const size_t len, str_buf_t* co
             {
                 return false;
             }
+            if (0 == ch_val)
+            {
+                return false;
+            }
             str_buf_printf(p_dst, "%c", (char)ch_val);
             p_cur += URL_ENCODE_NUM_CHARS_PER_BYTE;
         }
@@ -142,6 +146,7 @@ url_n_decode_with_alloc(const char* const p_src, const size_t len)
     }
     if (!url_n_decode_to_str_buf(p_src, len, &str_buf))
     {
+        str_buf_free_buf(&str_buf);
         return str_buf_init_null();
     }
     return str_buf;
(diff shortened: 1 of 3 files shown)


ojousima (approved):
Thanks

Issue #1213, opened 2026-03-23
CRA: 303645 IXIT-10-SecParam


ojousima:
## IXIT 10-SecParam: Security Parameters

The completed IXIT lists all sensitive (public and critical) security parameters that are persistently stored on the DUT during intended usage. The pro forma contains the following entries and is typically filled out in form of a table.

| Field | Description |
|-------|-------------|
| **ID** | Unique per IXIT identifier, assigned using a sequential numbering scheme or other labelling scheme. _Example: `SecParam-1` or `SecParam-Pswd`_ |
| **Description** | Brief description of the security parameter, including its purpose. Indicates whether the parameter is a hard-coded unique per-device identity used for security purposes and/or hard-coded in device software source code. |
| **Type** | Indication whether the security parameter is **public** or **critical**. (Defined in ETSI TS 103 645 / ETSI EN 303 645.) |
| **Security Guarantees** | Description of the realized baseline security objectives and threats the security parameter is protected against during persistent storage. |
| **Protection Scheme** | Description of the measures applied to achieve the Security Guarantees, including the principals and roles through which access to the parameter is possible, and the privileges associated with each role. |
| **Provisioning Mechanism** | _(Critical parameters only)_ Description of the mechanism through which the parameter is assigned its value for DUT operation. Assignment can happen during initialization or in initialized state (e.g. when a device functionality relying on the parameter is activated by the user). Possible mechanisms include persistent configuration data, runtime configuration data, protocol negotiation, and assignment to a default value. |
| **Secure Communication Mechanisms** | Reference to secure communication mechanisms in IXIT 11-SecComMech used to secure the communication of the parameter, and an indication whether the communication is done via remotely accessible interfaces. |
| **Generation Mechanism** | _(Critical parameters only, if used for integrity/authenticity checks of software updates or for protection of communication with associated services)_ Description of the mechanism used to generate parameter values, with an indication of the parameter's use. _Example: References to a standard random number generator and applicable design documents._ |

PR #1541, opened 2026-08-21
docs(cra): #1213 update ixit_10-SecParam.md with physical security notes and NVS encryption roadmap

TheSomeMan:


ojousima (approved):
Thanks

